PR #3, opened 2022-07-10
Rename italian translation to Localizable.strings and putting it into it.lproj

ZhichGaming:


lixin9311:
There are missing translations. But I will approve the change for now.

PR #10, opened 2022-07-12
Updated the Japanese localisations.

voltaicspy:
I also might have accidentally made 2 commits haha... I only fixed the Japanese ones and changed a bit of the English sentences for better understanding


lixin9311 (changes requested):
I think the translation you patched is from Google maybe?

これ、グーグル翻訳なの？www

PR #27, opened 2022-07-14
Added auto-migrate function for automating app migration process

Wind-Explorer:
and also modded some button placement and choice of words for error message regarding how COD and PUBG wont work

Issue #14, opened 2022-07-12
Migration Feature


ZhichGaming:
Add the option to migrate the files from .me to .io when opening for first time.

Depal1:
This feature won’t be implemented. Read https://github.com/PlayCover/PlayCover/pull/40 for more information. Instructions for a migration that consists of deleting apps installed pre 1.0.6 (just to free up space) and reinstalling them might need to be included on the documentation.

ZhichGaming:
I thought that symlink will not be implemented since it would not work if the original PlayCover is completely deleted? As said by lixin9311

> #Since we have changed the installation path of PlayTools https://github.com/PlayCover/PlayCover/blob/develop/PlayCover/Utils/PlayTools.swift#L101-L106

I don't think we can consider the Apps from the original PlayCover to be compatible with the new community version.

Once you completely delete the original PlayCover, the app installed by it will no longer be able to run.

Thus, I don't encourage any in-place migrations.

PR #38, opened 2022-07-14
playtools: follow Container symlink

hayatofujii:
To avoid reinstalling PlayCover'ed apps, I symlinked Containers/io.PlayCover.PlayCover to Containers/me.PlayCover/PlayCover. But swift doesn't honor those, so there is an ugly hack to support it.


Depal1:
Related to Pull request #27 and Issue #14.

PR #40, opened 2022-07-15
playtools: follow Container symlink

hayatofujii:
To avoid reinstalling PlayCover'ed apps, I symlinked Containers/io.PlayCover.PlayCover to Containers/me.PlayCover/PlayCover. But swift doesn't honor those, so there is an ugly hack to support it.

Did stupid renaming branches..., so, reopening it with the right refs.

Anyway, as noted by @Depal1, this is related to PR #27 and Issue #14


lixin9311:
Thanks for the PR.

Since we have changed the installation path of PlayTools https://github.com/PlayCover/PlayCover/blob/develop/PlayCover/Utils/PlayTools.swift#L101-L106
I don't think we can consider the Apps from the original PlayCover to be compatible with the new community version.

Once you completely delete the original PlayCover, the app installed by it will no longer be able to run.

Thus, I don't encourage any in-place migrations.

hayatofujii:
Fair enough. Did this change just because I was doing some regression tests and I was pretty annoyed by that. Closing up the PR, but the patchset will be still alive.

PR #49, opened 2022-07-16
Implement Discord Activity(Rich Presence)

KhoraLee:
This PR implement discord activity(also known as Rich Presence) feature.
When you open app from PlayCover, https://github.com/PlayCover/PlayTools/pull/2 will connect to discord client via IPC and update activity.

Default activity looks like below.
|image|Default Activity|
|---|---|

Because discord does not allow us to change name of application, It will display name as `PlayCover` and show application name(from bundle) in detail field.

**But** user can set specific application name by creating their own application in [Discord Developer Portal](https://discord.com/developers/applications). Also user can change details, state string and image shown at left side.

By setting custom activity like below, user can deliver discord activity as their wish.
image|스크린샷 2022-07-16 오후 5 44 56image|스크린샷 2022-07-16 오후 5 45 15|
|---|---|---|

### DIFF
--- a/PlayCover.xcodeproj/project.pbxproj
+++ b/PlayCover.xcodeproj/project.pbxproj
@@ -71,6 +71,7 @@
 		AA818CB5287ABEC3000BEE9D /* Yams in Frameworks */ = {isa = PBXBuildFile; productRef = AA818CB4287ABEC3000BEE9D /* Yams */; };
 		ABED59832887A32F004D782B /* MenuBarView.swift in Sources */ = {isa = PBXBuildFile; fileRef = ABED59822887A32F004D782B /* MenuBarView.swift */; };
 		B1084E1F28AA80F400E399FA /* AppSettingsVM.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1084E1E28AA80F400E399FA /* AppSettingsVM.swift */; };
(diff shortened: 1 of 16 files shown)


lixin9311:
This is coool 🔥

Depal1:
Hi! There was a merge with a lot of style fixes all over the code. Hence, a rebase is required.

KhoraLee:
@Depal1 
Rebased, thx.

IsaacMarovitz:
Text will need to be reworked to follow #86 locale style.

ZhichGaming:
Rich Presence does not work for me. I used the ID with the ID if found with right click on profile > Copy ID. I tried disabling and reenabling Enable Discord Activity but it still does not even show PlayCover(or the app launched) on my profile.  
Screen Shot 2022-08-24 at 21 44 55

KhoraLee:
@ZhichGaming 
If the `Carthage.resolved` file is not same with `github "KhoraLee/PlayTools" "b8d607a8da22a9ffbbde7006828db01cba73238a"` then do `carthage update --use-xcframeworks` and try again

ZhichGaming:
Thanks, it now works! And pretty quickly at that! I didn't even need to configure anything in the Discord Developer Portal except needing to create the application. I'd suggest that instead of ClientID, it should be Application ID since it's called that in the dev portal. People would get confused with User ID.
Screen Shot 2022-08-25 at 15 34 04

Depal1 (approved):

DarthMDev:
Why are some app defaults for the images genshin impact?

PR #51, opened 2022-07-16
Change the MacOS cursor behavior upon button hover

MrKai77:
This makes it more apparent to the end user that the button is a button.

https://user-images.githubusercontent.com/68963405/179370382-f5af3769-2188-48c8-816e-215e5034c177.mov


JoseMoreville:
Hi, buttons not always need to change cursor style, it's not a practice recommended by apple style rules.

Cursor pointer should be applied to hyperlinked text.
 [Apple human interface guidelines pointing devices ](https://developer.apple.com/design/human-interface-guidelines/inputs/pointing-devices/)

JoseMoreville (changes requested):
pointing hand Pointing hand should be used for links

MrKai77:
Alright, so would it make sense if:
• "Download more apps" changes the cursor since it opens a link to more .ipa files
• "Export to Sideloadly" changes the cursor since it opens Sideloadly
• "Add app" doesn't change the cursor since it prompts the user to open a file
• "Notices" doesn't change the cursor since it shows more information when clicked
• "Join Discord Server" changes the cursor since it opens Discord
• "Enable PlaySign" changes the cursor since it opens a "document" with instructions
• "Update app" changes the cursor since it opens the browser for a new version of the app
• "Crash" in debugging mode doesn't change the cursor since it doesn't open anything
• The "PlayAppView" function changes the cursor since it opens the clicked installed app
• The "StoreAppView" function changes the cursor since it downloads the .ipa file in the browser
Let me know if I missed anything!

JoseMoreville:
Is normal to think that way and valid, yet they are buttons, I'm not going to say that I follow this rule all the time, but due this being an open source app, i'd like to follow a guideline. Even though they do actions that might be considered. I'd like to point out that.

Buttons are not links, even if they can open a browser window. while links open pages and other resources, they are not meant to interact with the data changing it like buttons do.

If you check your mac UI you'll see that most of them have the arrow when you hover over them.

MrKai77:
Interesting, so would you recommend only having the "Join Discord Server" button change the cursor?

JoseMoreville:
> Interesting, so would you recommend only having the "Join Discord Server" button change the cursor?

Yes, since it is a hyperlink text I'd be nice to have it

MrKai77:
Alright, the "Join Discord Server" is now the only button that changes the cursor!

MrKai77:
Based on the new Better-UI commit, it looks like the Join Discord Server" button is non-existent, therefore making this pull request useless. As a result, I'll close this pull request.

Issue #52, opened 2022-07-16
Install xTools button's color is not uniform with other buttons. 


ZhichGaming:
It is not based on accent color.
<img width="1094" alt="Screen_Shot_2022-07-16_at_4 23 28_PM" src="https://user-images.githubusercontent.com/93055907/179371016-d92bbd37-334e-45ce-b273-d8d8c33fbd2a.png">

JoseMoreville:
Thank you, will check it and fix it :)

PR #54, opened 2022-07-17
style: :art: xcode CLI install button style is now consistent

JoseMoreville:
* Added more log info for app path
Issue #52


JoseMoreville:
Thanks to @ZhichGaming for the letting us know the problem

PR #65, opened 2022-07-19
UI Updates

lockieluke:
Updated title bar to Big Sur's style and App Settings buttons

image

image

### DIFF
--- a/PlayCover/View/AppSettingsView.swift
+++ b/PlayCover/View/AppSettingsView.swift
@@ -74,8 +74,8 @@ struct AppSettingsView : View {
                 Button("Reset settings and keymapping"){
                     resetedAlert.toggle()
                     settings.reset()
-                }.buttonStyle(CancelButton()).padding()
-                Button("Ok"){
+                }.buttonStyle(.bordered).padding()
+                Button("Done"){
                     settings.keymapping = keymapping
                     settings.adaptiveDisplay = adaptiveDisplay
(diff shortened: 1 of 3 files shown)


ohaiibuzzle:
I am not too sure if this is very necessary

Regarding the title bar, the HIG says this: "Display a title unless the window’s content provides enough context to make one unnecessary. For document windows, the title is the name of the document or Untitled (for new documents). For app windows, the title is the name of the app", which I think PlayCover qualifies for

As for the buttons, maybe?
I think we should have more opinions on this

lockieluke:
Should probably move the search bar and the buttons to the title bar

zanderp25:
I don't like how the done button is a link, either.

ohaiibuzzle:
> Should probably move the search bar and the buttons to the title bar

We may end up making the buttons menu items at some point as they are very much optional, and may end up rarely used (the show app links option already is in the latest build).
That leaves the search bar, which, probably better off just spanning the entire window than constraint to the top right.

lockieluke:
that's what I meant we should put the search field into the NSToolbae

Depal1:
As stated above, the titlebar should not be included, and the 'Done' button would be better off bordered and with accent color to follow macOS' design.

lockieluke:
Ok got it I'll try to implement those changes

Depal1:
Superseded by https://github.com/PlayCover/PlayCover/pull/135.

PR #69, opened 2022-07-20
fix: disable tabs in PlayCover

ohaiibuzzle:
Tabs in macOS are normally automatically enabled (accessible by pressing Shift + Command + \\)
Since this function is practically useless for PlayCover, we should disable it


ZhichGaming:
I know that this comment has no use, (pls dont ban me) but we are already at our 69th pull request! Nice!

Depal1 (approved):

PR #70, opened 2022-07-20
fixed simplified chinese + updates to pc ver1.0.6

JailbreakGentleman:
fixed some of the wording and added some coordinating conjunctions so the app is more appealing and understandable to new users. (keep in mind, im new to github.)


IsaacMarovitz:
Will need to be rebased to follow #86 locale style.

PR #71, opened 2022-07-20
Move the app links checkbox into the View menu

ohaiibuzzle:
Moved the app links checkbox into the View Menu and removed the checkbox on the main window to free up space for the search bar
viewmenu


Depal1 (changes requested):
Thanks for the comments! They have to go before merging though.

Depal1 (approved):

PR #81, opened 2022-07-21
Remove deprecated and unused Objective-C services

IsaacMarovitz:
Removed `LaunchServicesWrapper`, which uses a deprecated API to set `.ipa`s to open with PlayCover by default without alerting the user.

Removed `LoginKitWrapper`, which uses a deprecated API to manage login items, and is unused.

Fixed a deprecation warning.


IsaacMarovitz:
Rebasing and resubmitting bare with me 💀

PlayCover/View/PlayCoverApp.swift
```diff
@@ -31,7 +31,6 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 		UserDefaults.standard.register(
 			defaults: ["NSApplicationCrashOnExceptions" : true]
 		)
-        LaunchServicesWrapper.setMyselfAsDefaultApplicationForFileExtension("ipa")
```
zanderp25, line 34: Doesn't it need this to set itself as a default for ipa files, or is it done somewhere else?

IsaacMarovitz: It was Candy's opinion that setting PlayCover to open `.ipa`s by default without alerting the user isn't great practice. Elsewhere in the `Info.plist` it specifies that PlayCover can open `.ipa` files so it should still be available as an option, but just wont be set as default.

zanderp25: Ah, ok. Makes sense.

IsaacMarovitz:
Removed animation deprecation fix. Now part of #96.

Depal1 (approved):

PR #88, opened 2022-07-22
fix: Disable swiftlint on build actions and create SwiftLint action

Candygoblen123:
this fixes the current build action, and adds a new SwiftLint action

The SwiftLint action runs on PRs and on push


Candygoblen123:
Let me know if you would like me to make the lint action strict, so that it will always fail on warn.  This will make linting warnings more prevalent in Github

Depal1 (approved):

Issue #85, opened 2022-07-22
Prevent display from turning off while playing


kamui545:
It would be nice if there is any way to prevent display from turning off while playing.

I have very strict settings: 1-2 min before showing screen saver to avoid screen burn.
The screen saver often shows up during dialogues and cut-scenes.

Thank you!

ohaiibuzzle:
Will try to implement it. In the meantime, you can run `caffeinate` in your terminal when you launch your game which will disable all power saving features

PR #95, opened 2022-07-24
[May be breaking] Adding option to disable display timeout per-app

ohaiibuzzle:
This feature was suggested in issue #85 

This gives the user an option to completely disable display sleep when a specific app is running. However, because of this, we had to change to using `NSWorkspace.shared.openApplication` to give PlayCover a handler to the running app and enable/disable display sleep accordingly, so this PR is may have unintended issues.

### DIFF
--- a/PlayCover/View/PlayAppView.swift
+++ b/PlayCover/View/PlayAppView.swift
@@ -72,6 +72,12 @@ struct PlayAppView: View {
                     Image(systemName: "folder")
                 })
                 Button(action: {
+                    app.settings.disableTimeout.toggle()
+                }, label: {
+                    Text(app.settings.disableTimeout ? "Enable Display Sleep" : "Disable Display Sleep")
+                    Image(systemName: "display")
+                })
+                Button(action: {
                     app.openAppCache()
                 }, label: {
                     Text("Open app cache")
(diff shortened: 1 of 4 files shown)


PlayCover/View/PlayAppView.swift
```diff
@@ -64,7 +64,12 @@ struct PlayAppView: View {
                     Text("Show in Finder")
                     Image(systemName: "folder")
                 })
-
+                Button(action: {
```
Candygoblen123, line 67: It would be more appropriate to use a checkbox `Toggle` here instead of a button.

Candygoblen123 (changes requested):

PlayCover/View/PlayAppView.swift
ohaiibuzzle: > It would be more appropriate to use a checkbox `Toggle` here instead of a button.

Simple to explain. First of all I don't do that because it would involve creating a new state and due to the way the app menu is created, we'll have to do all kinds of funny business just to change it to a toggle

Second is that this happens if you use a Toggle.

image

image

Not ideal.

Candygoblen123: Oh, this is in the right click menu. It would be better to put it in the App Settings View instead

Depal1 (approved):

PlayCover/View/PlayAppView.swift
ohaiibuzzle: Moved to App Settings. Someone need to clean that thing up though. We are already running at the 10 views limit

Depal1 (approved):

PR #120, opened 2022-07-31
fix: Update account switcher UI

Candygoblen123:
- require all fields to be filled out before the store button becomes available
- add "Exit" button to store account sheet
- switch from using pill-shaped buttons to using system-like buttons
- add confirmation and warning when deleting an account 
- add confirmation and warning that restoring an account will override the currently signed-in account
- fixed a force try crash when users tried to store an account without opening the game for the first time
- removed various `swiftlint:disable` comments that didn't make much sense, and fixed the listing errors

### DIFF
--- a/PlayCover/View/ChangeGenshinAccountView.swift
+++ b/PlayCover/View/ChangeGenshinAccountView.swift
@@ -10,8 +10,7 @@ struct ChangeGenshinAccountView: View {
     @Environment(\.presentationMode) var presentationMode
     @State var folderName: String = ""
     @State var accountList: [String] = getAccountList()
-    // swiftlint:disable multiple_closures_with_trailing_closure
-    // swiftlint:disable opening_brace
+    @State var restoreAlert: Bool = false
     var body: some View {
         VStack(alignment: .center, spacing: 16) {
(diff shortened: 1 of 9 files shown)


PlayCover/View/ChangeGenshinAccountView.swift
```diff
@@ -21,28 +20,34 @@ struct ChangeGenshinAccountView: View {
...
+                    }, label: {
                         HStack {
                             Image(systemName: "person.fill")
                             Text(account)
                         }.frame(minWidth: 300, alignment: .center)
-                    }.controlSize(.large).buttonStyle(GrowingButton()).font(.title3)
+                    }).controlSize(.large).buttonStyle(.automatic).font(.title3)
+                        .foregroundColor(.accentColor)
                         .frame(width: 300, alignment: .center)
+                        .alert("Really Restore Account?", isPresented: $restoreAlert, actions: {
```
JoseMoreville, line 32: I'm not really sure about the message on this alert. Otherwise everything works great. Thank you Candy!

JoseMoreville (approved):

PR #129, opened 2022-08-01
[enhancement] Let user know upon install about overestimate of Xcode CLI Tool install prompt

Wind-Explorer:
Added a handy little button right below the "Install" button in the install Xcode CLI Tools page which brings the user to one of Tom Scott's video on why the install prompt say 100+ hours remaining.

### DIFF
--- a/PlayCover/View/AppSettingsView.swift
+++ b/PlayCover/View/AppSettingsView.swift
@@ -106,7 +106,7 @@ struct AppSettingsView: View {
                 Button("settings.reset") {
                     resetCompletedAlert.toggle()
                     settings.reset()
-                }.buttonStyle(GrowingButton()).padding()
+                }.controlSize(.large).padding()
                 Button("button.OK") {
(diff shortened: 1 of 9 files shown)


Depal1:
Is it really necessary to bundle a video for that? The app would not look so clean.

Wind-Explorer:
> Is it really necessary to bundle a video for that? The app would not look so clean.

There are more than the normal amount of people that have been questioning the time estimate from the prompt in Discord, expressing their worries, and hence I feel like we can include some form of indication to let the user know that it is normal and the user should calm down and wait patiently for the install to finish instead of jumping straight onto Discord and send screenshots asking why the need for 120 hours. On the other hand, the link only appears when PlayCover is asking for installing Xcode CLI Tools. The button is otherwise never shown anywhere else.

Depal1:
Personally I would rather say that behaviour is expected instead of linking a video.

PR #133, opened 2022-08-01
fix: Fix infinity loop in SetupView's alert modal

KhoraLee:
Fix infinity loop when entering mac password to enable PlaySign.
Fix showing raw key string and duplicate key string with different value. (Duplicate `setup.enterLoginPassword`)
SetupView Button will disappear right after Enabling PlaySign.

Currently, after enabling play sign button will remain and if you click button app does not response and cmd+Q to quit is not work. And there's no way to cancel entering password in alert modal because it keep asking for password with new alert modal

### DIFF
--- a/PlayCover/Utils/Shell.swift
+++ b/PlayCover/Utils/Shell.swift
@@ -143,7 +143,7 @@ class Shell: ObservableObject {
         }
 
     @discardableResult
-    static func shell(_ command: String, print: Bool = true) -> String {
+    static func shell(_ command: String, print: Bool = false) -> String {
         let task = Process()
         let pipe = Pipe()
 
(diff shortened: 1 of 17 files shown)


Candygoblen123:
Leaving the PlaySign button there Is intentional, since you have to restart your entire Mac for it to be truly enabled.

KhoraLee:
@Candygoblen123 After entering password to do nvram thing, if playsign button('Experiencing issue?') get clicked then app become unable to do anything. So I removed that button after showing 'need reboot alert'

https://user-images.githubusercontent.com/18005062/182332404-729f2956-5087-45f7-bb79-822365710960.mov

스크린샷 2022-08-02 오후 5 33 54

Candygoblen123:
Oh, my bad, I misunderstood what this PR was fixing. Since you've already created a UI state to remove the button, can you also add a "You need to reboot your Mac to continue" message similar to the Xcode Command Line Tools one.

KhoraLee:
This will shown after entering password to enable PlaySign
image

Depal1 (approved):

PR #137, opened 2022-08-02
chore: Add sparkle updater push to CI

Candygoblen123:
Please get in contact with me on discord so that I can securely send you the sparkle key.


lixin9311:
Hi @Candygoblen123 , thanks for the PR.
The workflow puts the build artifact directly to the website repo, which is an anti-pattern.
I would like to put the URL of the download link from the Github release directly into the appcast file.
Then we don't need to store the binary files in git repos.

Candygoblen123:
> I would like to put the URL of the download link from the Github release directly into the appcast file.

Doing this would not only significantly increase the complexity of the workflow, but would also fragment where our releases are stored, since we do not create GH releases for nightlies.

PR #150, opened 2022-08-12
feature: ✨ Now you can switch between Ipad models on app settings.

JoseMoreville:
Like the title says, this PR aims to add a picker on app settings to let the user select an ipad pro model to run the app on.
The models I added are Ipad pro 1st, 3rd and 5th gen. I plan to add iphone later but I still have to think in a way to avoid problems with this.

This PR needs https://github.com/PlayCover/PlayTools/pull/13 to work

### DIFF
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -191,6 +191,17 @@ class AppSettings {
             dictionary = dict
         }
     }
+    private static let ipadModel = "pc.ipadModel"
+    var ipadModel: String {
+        get {
+            dictionary[AppSettings.ipadModel] as? String ?? "iPad8,6"
+        }
+        set {
+            var dict = dictionary
+            dict[AppSettings.ipadModel] = newValue
+            dictionary = dict
+        }
+    }
 
     private var allPrefs: [String: Any] {
         get {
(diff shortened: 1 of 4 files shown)


JoseMoreville:
Please, do not merge it until there's a new version of playtools

Depal1:
I would recommend adding more detailed information on the models in question, if not on this PR, on a future one (particularly as a dropdown menu). This need arises since not all first generation iPad Pro models share the same onboard memory specifications.

The models in use and their data are left for future reference.
iPad6,7 - iPad Pro (12.9-inch) (1st generation) - A9X - 4Gb
iPad8,6 - iPad Pro (12.9-inch) (3rd generation) - A12Z - 4Gb (6Gb?)*
iPad13,8 - iPad Pro (12.9-inch) (5th generation) - M1 - 8Gb (16Gb?)*

*These models come with more memory on high storage configurations. Since they both share the same model identifier, it is unbeknownst to me if these higher end configurations are in use on PlayCover, or the former ones are.

ohaiibuzzle:
I wonder if it is possible to have a few iPad presets, then an option to input your own Model and Motherboard Identifier. Looking at the changes made to PlayTools, this may takes a bit more logic changes, but would effectively future proof this feature (in case someone wants to go back to an earlier version of PC)

JoseMoreville:
> I wonder if it is possible to have a few iPad presets, then an option to input your own Model and Motherboard Identifier. Looking at the changes made to PlayTools, this may takes a bit more logic changes, but would effectively future proof this feature (in case someone wants to go back to an earlier version of PC)

This will be done in a future, not sure when i'm going to start working on it but its planned as said on discrod

Depal1 (approved):

PR #158, opened 2022-08-14
[Enhancement] Improved device picker UI

Wind-Explorer:
Interface for other locales were broken without the new keys added for each locale hence this PR solves the issue as well as modifying picker style for future-proof

### DIFF
--- a/PlayCover/de.lproj/Localizable.strings
+++ b/PlayCover/de.lproj/Localizable.strings
@@ -11,6 +11,7 @@
 "playapp.storeCurrentAccount" = "Store current account";
 "playapp.activateAccount" = "Activate an account";
 "playapp.deleteAccount" = "Delete an account";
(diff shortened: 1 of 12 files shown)


PlayCover/de.lproj/Localizable.strings
```diff
@@ -85,6 +86,7 @@
 
 "settings.picker.displayRefreshRate" = "Screen refresh rate";
 "settings.picker.screenSize" = "Display resolution";
+"settings.picker.ipadModel" = "Select iPad Pro model";
```
Depal1, line 89: I would recommend changing this locale to "settings.picker.iosDevice" or something similar since we may add iPhones under the same system in the future.

Depal1 (changes requested):

Depal1 (approved):

PR #225, opened 2022-08-21
[Enhancement] Make settings view scrollable

Wind-Explorer:
Settings view now takes up way less space thanks to the implementation of Scroll View


Candygoblen123:
This seems like more of a duck tape fox than a proper fix. Seeing as a proper fix with a tabbed view is in the works, I cannot approve this is good faith.

PR #231, opened 2022-08-22
Translation section in README

zanderp25:
I looked into the api and it seemed like weblate would not let it retrieve the data... either that or they changed the Weblate API and the other site hasnt updated yet.

Also, I added a translations section, which might seem more fitting, but I'm not sure if i want to keep that wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 [![Stargazers][stars-shield]][stars-url]
 [![Issues][issues-shield]][issues-url]
 [![GPLv3 License][license-shield]][license-url]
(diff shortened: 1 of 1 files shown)


IsaacMarovitz:
The translation badge works some of the time (like right now)

IsaacMarovitz:
image
I'm guessing it's an issue on weblate's side

zanderp25:
Oh, well that's weird. It's still not working for me.

Screen Shot 2022-08-22 at 11 25 34

Should we still keep the translation section though?

zanderp25 (commented):
wait why doesn't this work

Candygoblen123:
seems to work for me

zanderp25:
Aight im gonna close this

PR #232, opened 2022-08-22
Fix all the stuff i didnt get done in 2.0.0 ui lol

IsaacMarovitz:
rainbows-cockroach
*Fixing TODOs...*
**Very Serious Changelog:**
- [x] Change app settings from a monolithic file to a per-app file stored in the Containers folder
- [x] Updated app settings to use a `Codable` struct for significantly better code readability
- [x] Fixed the appearance of `AsyncImage` and download icon from App Store for up-to-date icons and correct image size
- [x] Change keymapping so they are stored in the Containers folder and use `Codable`
- [x] Make 'reset settings' and 'reset keymapping' buttons functional
- [x] Finish localising strings

### DIFF
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -68,8 +68,7 @@ class AppSettings {
     public func decode() -> Bool {
         do {
             let data = try Data(contentsOf: settingsUrl)
-            let decoder = PropertyListDecoder()
-            settings = try decoder.decode(AppSettingsData.self, from: data)
+            settings = try PropertyListDecoder().decode(AppSettingsData.self, from: data)
             return true
         } catch {
             print(error)
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -42,7 +42,7 @@ class AppSettings {
     var container: AppContainer?
     var settings: AppSettingsData {
         didSet {
-            _ = encode()
+            encode()
         }
     }
 
@@ -52,7 +52,7 @@ class AppSettings {
         self.settingsUrl = AppSettings.appSettingsDir.appendingPathComponent("\(info.bundleIdentifier).plist")
         self.settings = AppSettingsData()
         if !decode() {
-            _ = encode()
+            encode()
         }
     }
 
@@ -64,6 +64,7 @@ class AppSettings {
         settings = AppSettingsData()
     }
 
+    @discardableResult
     public func decode() -> Bool {
         do {
             let data = try Data(contentsOf: settingsUrl)
@@ -76,6 +77,7 @@ class AppSettings {
         }
     }
 
+    @discardableResult
     public func encode() -> Bool {
         let encoder = PropertyListEncoder()
         encoder.outputFormat = .xml
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -21,6 +21,7 @@ struct AppSettingsData: Codable {
     var aspectRatio: Int = 1
     var notch: Bool = NSScreen.hasNotch()
     var bypass: Bool = false
+    var version: String = "2.0.0"
 }
 
 class AppSettings {
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -7,287 +7,86 @@ import Foundation
 import UniformTypeIdentifiers
 import AppKit
 
-class AppSettings {
-    // TODO: Use per-app files instead of a monolithic file
-    public static let settingsUrl = URL(fileURLWithPath: "/Users/\(NSUserName())/Library/Preferences/playcover.plist")
+struct AppSettingsData: Codable {
+    var keymapping: Bool = true
+    var mouseMapping: Bool = true
+    var sensitivity: Float = 50
 
-    public func sync() {
-        notch = NSScreen.hasNotch()
+    var disableTimeout: Bool = false
+    var iosDeviceModel: String = "iPad8,6"
+    var refreshRate: Int = 60
+    var windowWidth: Int = 1920
+    var windowHeight: Int = 1080
+    var resolution: Int = 2
+    var aspectRatio: Int = 1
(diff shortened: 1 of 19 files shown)


Depal1:
Could you please list what UI parts are you addressing?

Candygoblen123:
> rainbows-cockroach
> _Fixing TODOs..._

please be professional in your PR descriptions

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -67,33 +117,73 @@ struct StoreAppConditionalView: View {
...
+
+    public func getIconURLFromBundleIdentifier(_ bundleIdentifier: String,
+                                               _ region: StoreAppData.Region,
+                                               completion: @escaping (URL) -> Void) {
+        let url: URL
+
+        if region == .CN {
+            url = URL(string: "http://itunes.apple.com/lookup?bundleId=\(bundleIdentifier)" + "&country=cn")!
+        } else {
+            url = URL(string: "http://itunes.apple.com/lookup?bundleId=\(bundleIdentifier)")!
```
KhoraLee, line 111: If we don't specify country then it will query US store front. So all app not in US store will return empty.

PlayCover/Model/AppSettings.swift
```diff
@@ -7,287 +7,86 @@ import Foundation
...
-        set {
-            var prefs = allPrefs
-            prefs[info.bundleIdentifier] = newValue
-            allPrefs = prefs
+    let info: AppInfo
+    let settingsUrl: URL
+    var container: AppContainer?
+    var settings: AppSettingsData {
+        didSet {
+            _ = encode()
```
KhoraLee: I think using `@discardableResult` to suppress warning is better. (https://github.com/PlayCover/PlayCover/pull/178)

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -67,33 +117,73 @@ struct StoreAppConditionalView: View {
...
+                getIconURLFromBundleIdentifier(app.id, app.region) { url in
+                    iconUrl = url
+                }
+            }
+        }
+    }
+
+    public func getIconURLFromBundleIdentifier(_ bundleIdentifier: String,
+                                               _ region: StoreAppData.Region,
+                                               completion: @escaping (URL) -> Void) {
```
KhoraLee: Why not use async/await rather than completion handler

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -7,6 +7,57 @@
...
+    var languageCodesISO2A: [String]
+    var fileSizeBytes: String
+    var releaseDate: String
+    var artistId: Int
+    var artistName: String
+    var genres: [String]
+    var price: Float
+    var version: String
+    var wrapperType: String
+    var userRatingCount: Int
```
KhoraLee: Do we need all of these? Comment out unused ones.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -37,28 +88,27 @@ struct StoreAppView: View {
 struct StoreAppConditionalView: View {
     @State var app: StoreAppData
     @State var isList: Bool
+    @State var iconUrl: URL = URL(string: "https://google.com")!
```
KhoraLee: AsyncImage doesn't require `URL`. It can get `URL?` as parameter.

PlayCover/Views/App Views/StoreAppView.swift
IsaacMarovitz: This data may be useful in other places in the UI. I'll extract it to its own file for easier readability.

PlayCover/Views/App Views/StoreAppView.swift
IsaacMarovitz: Currently, the only regions defined by the `region` enum are Global and CN, so the only regions that need to be checked are those two. If in future we change to include more region identifiers, this will need to be altered to reflect that.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -67,33 +66,64 @@ struct StoreAppConditionalView: View {
...
             }
             .frame(width: 150, height: 150)
+            .task {
+                iconUrl = try? await getIconURLFromBundleIdentifier(app.id, app.region)
+            }
+        }
+    }
+
+    func getIconURLFromBundleIdentifier(_ bundleIdentifier: String,
+                                        _ region: StoreAppData.Region) async throws -> URL? {
```
KhoraLee: You are using do - catch statement inside function. So no need to make function throws.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -67,33 +66,64 @@ struct StoreAppConditionalView: View {
...
+            let decoder = JSONDecoder()
+            let jsonResult: ITunesResponse = try decoder.decode(ITunesResponse.self, from: data)
+            if jsonResult.resultCount > 0 {
+                return URL(string: jsonResult.results[0].artworkUrl512)!
+            } else {
+                return nil
+            }
+        } catch {
+            Log.shared.error("error: \(error)")
+            return nil
```
KhoraLee: Just return nil at the end of function.

PlayCover/Views/App Views/StoreAppView.swift
IsaacMarovitz: URLSession can also throw errors, so the throw is necessary.

KhoraLee: Move URLSession inside do-catch

PlayCover/Views/App Views/StoreAppView.swift
Depal1: There are several other regions that need to be added as exemplified in https://ipa.playcover.workers.dev/1:/IPAs/

IsaacMarovitz: Will handle in a different pr

Candygoblen123 (approved):

PR #240, opened 2022-08-24
Update README with iVoider's Github profile link

amirsaam:


Candygoblen123:
This is a good PR, but I don’t think we are interested in giving more attention to iVoider than we have to.

JoseMoreville:
I think the same, we should leave the name but linking is not needed

PR #247, opened 2022-08-25
Change min main window width + small refactoring

castlele:
# Main fix

## Min width before fix

Screenshot 2022-08-25 at 19 39 23

## Min width after fix

Screenshot 2022-08-25 at 21 14 32

# Additional changes

* Added `Constants` struct that should holds global constants in the app;
* Refactor some places with constants from `Constants`.

### DIFF
--- a/PlayCover/Utils/GlobalConstants.swift
+++ b/PlayCover/Utils/GlobalConstants.swift
@@ -1,0 +1,20 @@
+//
+//  GlobalConstants.swift
+//  PlayCover
+//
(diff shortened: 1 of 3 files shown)


Candygoblen123:
Hi, we’re only taking UI related PRs on the 2.0-UI branch. I believe that it’s already fix there, but could you verify and refactor your PR if not?

castlele:
Hi, I've created new [PR](https://github.com/PlayCover/PlayCover/pull/258) to 2.0-UI branch

PR #303, opened 2022-09-06
Fix keymappings

TheMoonThatRises:
Fixes bug where keymappings are exported to blank slates. Also fixes the bug where draggableButtonModels are not able to be exported.

### DIFF
--- a/PlayCover/Model/Keymapping.swift
+++ b/PlayCover/Model/Keymapping.swift
@@ -10,9 +10,9 @@ import Foundation
 import UniformTypeIdentifiers
 
 struct KeyModelTransform: Codable {
-    var size: Float
-    var xCoord: Float
-    var yCoord: Float
+    var size: CGFloat
+    var xCoord: CGFloat
+    var yCoord: CGFloat
 }
 
 struct ButtonModel: Codable {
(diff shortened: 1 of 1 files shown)


IsaacMarovitz:
Could you provide some examples of times where the current exporter fasley exports blank states?

TheMoonThatRises:
I have tested this mainly on Genshin, and it is when I create a keymap within it and try to export it. In the keymapping directory it shows the correct plist, but when it is exported, it becomes blank (in the sense that there are no keys on it).

ZhichGaming:
Do you mean for the new button only? The draggable button one?

TheMoonThatRises:
None of the buttons are saved when they are exported.

TheMoonThatRises:
Also, the draggable button option doesn't even have a key in the plist.

ZhichGaming:
I can reproduce this issue on the lastest nightly.

IsaacMarovitz:
I wasn't aware of the CGFloat/Float mismatch between PlayTools and PlayCover code. These changes are necessary.

IsaacMarovitz (approved):

Depal1 (approved):

PR #314, opened 2022-09-07
Legacy Settings & Keymapping Converter

IsaacMarovitz:


PlayCover/Utils/LegacySettings.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    static func convertLegacyKeymapFile(_ from: URL) -> Keymap? {
+        var dictionary: [String: Any]
+
+        do {
+            if let settings = try [String: Any].read(from) {
+                if !settings.isEmpty {
+                    dictionary = settings
+                    let legacyKeymaps = dictionary["pc.layout"] as? [String: Any] ?? [:]
+                    var keymap = Keymap(bundleIdentifier: "")
```
TheMoonThatRises: You can get the bundleIdentifier from
```suggest
from.deletingPathExtension().lastPathComponent
```

PlayCover/Utils/LegacySettings.swift
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    static func convertLegacySettingsDict(_ from: [String: Any]) -> AppSettingsData? {
+        var dictionary: [String: Any]
+
+        if !from.isEmpty {
+            dictionary = from
+
+            var settingsData = AppSettingsData()
```
TheMoonThatRises, line 65: The legacy keymap settings also has:
```suggest
pc.bypass, pc.enableWindowAutoSize, pc.gameWindowSizeHeight, pc.gameWindowSizeWidth, pc.ipadModel
```
which are stored in AppSettingsData

ZhichGaming:
This error should have a more detailed description such as: Legacy keymapping file names must be the same as the application bundle ID to be imported. Other than that, I did not run into any major issues. 

image

PlayCover/Utils/LegacySettings.swift
JoseMoreville: Those are not keymapping, those are settings that can be changed in menu of they're still available.

Depal1: This PR is aiming to preserve settings too.

IsaacMarovitz: Here's my rational for excluding those parameters.

`pc.bypass`: Jailbreak Bypass did not work as intended in the majority of legacy PlayCover
`pc.enableWindowAutoSize`, `pc.gameWindowSizeHeight`, `pc.gameWindowSizeWidth`: The way that resolution behaves in 2.0 has been almost entirely refactored. Setting new settings based on this data might cause unintended behaviour.
`pc.ipadModel`: M1 Model Identifier <2.0 was broken, now that has been fixed, the default has been changed to M1

JoseMoreville (approved):

PR #320, opened 2022-09-08
feat: New App Signing configuration experience.

ohaiibuzzle:
This commit brings a new, guided approach for configuring app signing (which make login in apps like Genshin works)

The "wizard" guides the user through disabling SIP and disabling AMFI in a more visual manner.

Here is what it looks like:
Menu accessible via PlayCover > Configure Signing:
image

AMFI:
out

Signing ready:
image

### DIFF
--- a/PlayCover.xcodeproj/project.pbxproj
+++ b/PlayCover.xcodeproj/project.pbxproj
@@ -68,6 +68,7 @@
 		AA71964B287A0EA800623C15 /* PlayRules.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA71964A287A0EA800623C15 /* PlayRules.swift */; };
 		AA71964D287A35F400623C15 /* com.nexon.bluearchive.yaml in Resources */ = {isa = PBXBuildFile; fileRef = AA71964C287A35F400623C15 /* com.nexon.bluearchive.yaml */; };
 		AA818CB5287ABEC3000BEE9D /* Yams in Frameworks */ = {isa = PBXBuildFile; productRef = AA818CB4287ABEC3000BEE9D /* Yams */; };
+		AB671B8C28C9C7DB00B7B102 /* SigningSetupView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AB671B8B28C9C7DB00B7B102 /* SigningSetupView.swift */; };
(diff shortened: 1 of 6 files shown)


ZhichGaming:
Can you insert a button that links to the docs? Do the question marks point to their instructions as to how to set them?

ohaiibuzzle:
> Can you insert a button that links to the docs? Do the question marks point to their instructions as to how to set them?

Those buttons provides information as to why we are disabling those macOS features when you hover over them
Docs, we have one on the Help menu, do you think we need it here as well?

amirsaam:
I think this gives a the user a feeling that he/she needs to do it once. more details i think is required under title.
like "you need to do this process everytime you want to login into a new account" or something like this.

ZhichGaming:
> > Can you insert a button that links to the docs? Do the question marks point to their instructions as to how to set them?
> 
> Those buttons provides information as to why we are disabling those macOS features when you hover over them Docs, we have one on the Help menu, do you think we need it here as well?

If It's only in the help menu, yes you should add a button linking to the docs maybe called "How to disable SIP and NVRAM" since people definitely would not know how to disable SIP without a guide and i would say most users do not know that we need to paste the nvram command in terminal.

ZhichGaming:
> I think this gives a the user a feeling that he/she needs to do it once. more details i think is required under title. like "you need to do this process everytime you want to login into a new account" or something like this.

That is too Genshin specific, that pop up applies for a lot of apps.

amirsaam:
> > I think this gives a the user a feeling that he/she needs to do it once. more details i think is required under title. like "you need to do this process everytime you want to login into a new account" or something like this.
> 
> That is too Genshin specific, that pop up applies for a lot of apps.

So, you don't have to disable sip everytime you want to login into a new account for a different app?

ZhichGaming:
I mean for some apps, you need SIP disabled the whole time not just for logging in.

amirsaam:
So there could simply be a note like: "not every app let's you stay logged in after enabling sip again" or something like this

ohaiibuzzle:
I mean we can put up an abomination to modern design, ala. a wall of text, like so
image
if you are so worried about that

amirsaam:
Lemme clarify, PlayCover dictionary is completely confusing. Words like PlayBook and PlayMap that currently are under use are not familiar to any user. I just saying we need to clarify not with a wall of words but with a couple of short notes.

Two note are necessary in my opinion that user should know:
- Some apps may be logged out after enabling SIP afterwards.
- If user wants to login into a new account, disabling SIP again is required.

Depal1 (approved):

PR #327, opened 2022-09-09
Code cleanup: removal of old resolution selector

Depal1:
This PR removes files of the former resolution selector that was used prior to the merge of https://github.com/PlayCover/PlayCover/pull/290.


IsaacMarovitz (approved):

IsaacMarovitz (approved):

JoseMoreville (approved):

JoseMoreville:
@Depal1 What do you think on creating a deprecated folder to keep code to remove in future versions. so we keep a better track of it.

Depal1:
I don’t think it is a good idea. Deprecated code will always be available for anyone interested on seeing it with version control, and easily accesible with tags, or even on a commit basis, as a core git feature. It is the recommended course of action as stated in git’s documentation.

PR #331, opened 2022-09-11
Fix automatic screen size detection

zanderp25:
Previous builds of PlayCover had an issue where the automatic screen size option would not use the correct aspect ratio (and apparently eventually became square?)

This fix sets the window size to the size of the full screen. If a notch is detected, it forces the aspect ratio to 16:10, which is the correct ratio of the screen size without the notch.

### DIFF
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -16,7 +16,7 @@ struct AppSettingsData: Codable {
     var iosDeviceModel = "iPad13,8"
     var windowWidth = 1920
     var windowHeight = 1080
-    var resolution = 2
+    var resolution = 1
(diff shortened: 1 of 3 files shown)


ZhichGaming:
Did you set display mode Auto as the default behavior?

zanderp25:
> Did you set display mode Auto as the default behavior?

I didn't change any default behavior

ZhichGaming:
I think it should default as auto since that is the ideal settings that will probably work for most users.

zanderp25:
one sec lemme fix the issues

zanderp25:
Alright the issues should be fixed now and i set the default

Depal1 (approved):

PR #332, opened 2022-09-11
General app settings tab

zanderp25:
I, and a few others, thought it would be nicer if the tabs with only a few items would be merged into a single "general" tab.
There are only a few changes needed with other localizations.


zanderp25:
ah yes, the error suppressor is erroring

zanderp25:
whoops did i accidentally commit my project file

ZhichGaming:
Could you attach some screenshots?

zanderp25:
Here's a screenshot
screenshot of general tab

Depal1:
This was discussed recently, but was rejected because the keymapping and jailbreak tabs will accommodate more settings in the short term, such as entitlement customization for the latter.

As for `import keymapping` and `export keymapping`, they are not really settings.

zanderp25:
> As for `import keymapping` and `export keymapping`, they are not really settings.

I forgot to mention that this was removed

PR #347, opened 2022-09-16
Reimplement `optool` in Swift

IsaacMarovitz:


Depal1:
When this is ready, it should be squashed before merging.

PlayCover/AppInstaller/Binary Patcher/PatchBinary.swift
```diff
@@ -0,0 +1,51 @@
...
+            return false
+        }
+    }
+
+    public static func removePlayToolsFrom(binaryURL: URL, dylibName: String) -> Bool {
+        do {
+            guard var binary = NSMutableData(contentsOf: binaryURL) else { return false }
+            var headers = Headers.headersFromBinary(binary: binary)
+
+            for index in 0...headers.count {
```
XuYicong: Does "..." operator include the ending bound? If so, it should be "..<" here. 
Or why don't say `for header in headers` directly?

PlayCover/AppInstaller/Binary Patcher/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+    private static func swap(_ bool: Bool, _ int: Int32) -> Int32 {
+        return bool ? Int32(CFSwapInt32(UInt32(int))) : int
+    }
+
+    private static func swap(_ bool: Bool, _ int: UInt32) -> UInt32 {
+        return bool ? CFSwapInt32(int) : int
+    }
+
+    public static func headerAtOffset(_ binary: NSData, _ offset: UInt32) -> thin_header {
+        let header = (binary.bytes + Int(offset)).load(as: mach_header.self)
```
XuYicong: In original C code, this was just `&binary.bytes[offset]` written in a convenient way. I believe there are simpler methods to do the same in swift. How about `binary.subdata()`?

PlayCover/AppInstaller/Binary Patcher/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+// swiftlint:disable type_name
+struct thin_header {
+    var offset: UInt32 = 0
+    var size: UInt32 = 0
+    var header: mach_header = mach_header()
+}
+
+class Headers {
+    public static func headersFromBinary(binary: NSData) -> [thin_header] {
+        var headers = Array(repeating: thin_header(), count: 4)
```
XuYicong: Declaring it's capacity is enough, no need to repeat that object four times.
But is 4 really the upper bound? I prefer a mutable array then append to it.

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+
+        let length: UInt = UInt(MemoryLayout.size(ofValue: dylib_command.self)) + UInt(dylibPath.count)
+        let padding: UInt = (8 - (length % 8))
+
+        let occupant = binary.subdata(with: NSRange(location: Int(macho.header.sizeofcmds + macho.offset + macho.size),
+                                                    length: Int(length + padding)))
+
+        if occupant[0] != 0 {
```
XuYicong, line 41: This was originally checking whether there are left padding bytes to insert a new load command. In this logic, this should check whether all bytes in this range are all 0. Merely checking the first byte is not enough.
Though I don't think the original `strcmp` checks all bytes. idk how it worked.

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+
+    @discardableResult
+    public static func removeLoadEntryFromBinary(_ binary: inout NSMutableData,
+                                                 _ macho: inout thin_header,
+                                                 _ payload: String) -> Bool {
+        var offset: Int = Int(macho.offset + macho.size)
+
+        var num: UInt32 = 0
+        var cumulativeSize: UInt32 = 0
+        var removedOrdinal: Int32 = -1
```
XuYicong: It was `uint32` in original C. -1 means the maximum value.

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+        dylib.current_version = 0
+        dylib.compatibility_version = 0
+        command.cmd = type
+        command.dylib = dylib
+        command.cmdsize = UInt32(length + padding)
+
+        var zeroByte: UInt = 0
+        let commandData = NSMutableData(data: Data())
+        commandData.append(&command, length: MemoryLayout.size(ofValue: dylib_command.self))
+        commandData.append(dylibPath.data(using: .ascii)!)
```
XuYicong: can it use UTF-8 here? Although original C code uses ASCII, this would fail once there are non-ASCII characters in the dylib path.

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+
+            let sizeData = Data(bytes: binary[offset+4..<offset+8].base
+                .bytes, count: 4)
+            let size = sizeData.withUnsafeBytes { $0.load(as: UInt32.self) }
+
+            switch cmd {
+            case LC_REEXPORT_DYLIB,
+            LC_LOAD_UPWARD_DYLIB,
+            LC_LOAD_WEAK_DYLIB,
+            UInt32(LC_LOAD_DYLIB):
```
XuYicong, line 105: why the heck aren't these constants beginning with `LC_` of the same type?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+                    print("Removing payload from \(LC(cmd))...")
+                    binary.replaceBytes(in: NSRange(location: offset,
+                                                    length: Int(size)), withBytes: nil, length: 0)
+                    num += 1
+                    cumulativeSize += size
+                    removedOrdinal = Int32(index)
+                }
+
+                offset += Int(size)
+            default:
```
XuYicong, line 122: have you missed a case branch?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+            case LC_REEXPORT_DYLIB,
+            LC_LOAD_UPWARD_DYLIB,
+            LC_LOAD_WEAK_DYLIB,
+            UInt32(LC_LOAD_DYLIB):
+                let command: dylib_command = (binary.bytes + offset).load(as: dylib_command.self)
+                let bytes = binary.subdata(with:
+                                            NSRange(location: offset + Int(command.dylib.name.offset),
+                                                    length: Int(command.cmdsize - command.dylib.name.offset)))
+                    .withUnsafeBytes({ $0.load(as: Array<UInt8>.self) })
+                let name = String(bytes: bytes, encoding: .utf8)
```
XuYicong, line 111: original C code does not specify the encoding here. Why is it utf8 here?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+                    binary.replaceBytes(in: NSRange(location: offset,
+                                                    length: Int(size)), withBytes: nil, length: 0)
+                    num += 1
+                    cumulativeSize += size
+                    removedOrdinal = Int32(index)
+                }
+
+                offset += Int(size)
+            default:
+                offset += Int(size)
```
XuYicong, line 123: is it necessary to put this `offset` back to `binary`?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    public static func binaryHasLoadCommandForDylib(_ binary: inout NSMutableData,
+                                                    _ dylib: String,
+                                                    _ lastOffset: inout UInt32,
+                                                    _ macho: thin_header) -> Bool {
+        var offset: Int = Int(macho.size + macho.offset)
+        var loadOffset = offset
+
+        for _ in 0...macho.header.ncmds {
```
XuYicong: may be `..<` instead of `...`.

XuYicong (commented):
Is it really fine to have a space in the directory name though?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
IsaacMarovitz: You have to specify the encoding in Swift

PlayCover/AppInstaller/Binary Patcher/Operations.swift
IsaacMarovitz: Ask Apple, these constants are defined in the MachO.loader framework

PlayCover/AppInstaller/Binary Patcher/Headers.swift
IsaacMarovitz: Swift arrays are always mutable

PlayCover/AppInstaller/Binary Patcher/Operations.swift
IsaacMarovitz: That case is only run in debug

PlayCover/AppInstaller/Binary Patcher/Operations.swift
XuYicong: yes but why utf8? why not ASCII or ISO-8859-1 or something else?

PlayCover/AppInstaller/Binary Patcher/Operations.swift
XuYicong: The `#ifdef DEBUG` macro only wraps a part of that case. The remaining part still runs even if not in debug.

Issue #304, opened 2022-09-06
[Bug]: Shared Executable Names between sagas causes Replacement


amirsaam:
### Describe the bug

Games that have a shared Executable Name with different Bundle Names (saga's mostly) will replace each other.

### Steps to reproduce

All of Infinity Blades are using "SwordGame" as Executable Name. So:

1. Install IB1 or 2 or 3
2. Try Installing different one from step 1
3. The one from step 2 will replace installed one in step 1

### Expected behaviour

PlayCover should be able to differentiate between such games and install them separately not replace them on each other.

Solution 1: Use Bundle Identifiers as .app names
Solution 2: While installing an app, check the executable name if it already exist, then check if bundle identifiers are different or not, if not send a replacement warning to user if they were different add something at the end of executable app name. (basically you will add a replace warning while fixing this bug)

### Crash log

```shell
None
```

### What version of PlayCover are you using?

Nightly/beta

### What version of macOS are you using?

Monterey (macOS 12)

### Issue Language

- [X] Yes my issue is written in English

ZhichGaming:
This is a feature request and not a bug.

amirsaam:
@ZhichGaming that's a bug if causes different apps replace each other, is it not?

ZhichGaming:
A feature request can be related to a bug and you can give details on it in the first box.

amirsaam:
@ZhichGaming I have sent feature requests before, but I think it's better to be described as a bug because it's not a problem on an actual iOS device.
iOS uses Bundle Name as Home Screen app name.

ZhichGaming:
Hmm ok, well it's not a big deal anyway.

amirsaam:
This issue has been fixed by side effects of PR #368 with commit `Change .app name to display name on install`

PR #368, opened 2022-09-21
Updated IPA Source Management

IsaacMarovitz:
https://user-images.githubusercontent.com/42140194/193436833-dc2819ca-d6fc-4d17-a657-5bae5b250d5a.mov

- Remove hardcoded `.json` both for flexability, and for copyright
- Allow users to add their own in-app IPA sources
- In-app downloads

**Other Changes/Bug Fixes:**
- `.app` folders will be renamed to match display name at install time. This means apps will now display with the correct name in the dock
- Updated the format of IPA Source `.json`s, an example of which can be found [here](https://raw.githubusercontent.com/IsaacMarovitz/PlayCover-IPA-Source-Tests/main/updatedsource.json)
- Fixed an issue where duplicate apps in the `Containers` folder with the same bundle ID would display incorrectly in grid view, but correctly in list view
- Fixed an issue where the display name of apps was sometimes equal to ""
- Cleaned up App Library and IPA Library views

### DIFF
--- a/PlayCover/AppInstaller/Installer.swift
+++ b/PlayCover/AppInstaller/Installer.swift
@@ -183,7 +183,11 @@ class Installer {
 
     /// Generates a wrapper bundle for an iOS app that allows it to be launched from Finder and other macOS UIs
     static func wrap(_ baseApp: BaseApp) throws -> URL {
-        let location = PlayTools.playCoverContainer.appendingPathComponent(baseApp.url.lastPathComponent)
+        let info = AppInfo(contentsOf: baseApp.url
+            .appendingPathComponent("Info.plist"))
+        let location = PlayTools.playCoverContainer
+            .appendingPathComponent(info.displayName)
+            .appendingPathExtension("app")
         if FileManager.default.fileExists(atPath: location.path) {
             try FileManager.default.removeItem(at: location)
         }
--- a/PlayCover/Views/App Views/StoreAppView.swift
+++ b/PlayCover/Views/App Views/StoreAppView.swift
@@ -14,21 +14,76 @@ struct StoreAppView: View {
 
     @State var app: StoreAppData
     @State var isList: Bool
+    @State var isDownloading: Bool = false
+    @State var downloadProgress: Double = 0
+    @State var observation: NSKeyValueObservation?
 
     var body: some View {
         StoreAppConditionalView(selectedBackgroundColor: $selectedBackgroundColor,
                                 selectedTextColor: $selectedTextColor,
                                 selected: $selected,
+                                isDownloading: $isDownloading,
+                                downloadProgress: $downloadProgress,
                                 app: app,
                                 isList: isList)
-            .gesture(TapGesture(count: 2).onEnded {
-                if let url = URL(string: app.link) {
-                    NSWorkspace.shared.open(url)
-                }
-            })
-            .simultaneousGesture(TapGesture().onEnded {
-                selected = app
-            })
+        .gesture(TapGesture(count: 2).onEnded {
+            if let url = URL(string: app.link) {
+                downloadApp(url)
+            }
+        })
+        .simultaneousGesture(TapGesture().onEnded {
+            selected = app
+        })
+    }
+
+    func downloadApp(_ url: URL) {
+        lazy var urlSession = URLSession(configuration: .default, delegate: nil, delegateQueue: nil)
+        let downloadTask = urlSession.downloadTask(with: url, completionHandler: { url, urlResponse, error in
+            observation?.invalidate()
+            downloadComplete(url, urlResponse, error)
+        })
+
+        observation = downloadTask.progress.observe(\.fractionCompleted) { progress, _ in
+            downloadProgress = progress.fractionCompleted
+        }
+
+        downloadTask.resume()
+        isDownloading = true
+    }
+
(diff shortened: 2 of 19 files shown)


PlayCover/AppInstaller/Installer.swift
```diff
@@ -183,7 +183,11 @@ class Installer {
 
     /// Generates a wrapper bundle for an iOS app that allows it to be launched from Finder and other macOS UIs
     static func wrap(_ baseApp: BaseApp) throws -> URL {
-        let location = PlayTools.playCoverContainer.appendingPathComponent(baseApp.url.lastPathComponent)
+        let info = AppInfo(contentsOf: baseApp.url
+            .appendingPathComponent("Info.plist"))
+        let location = PlayTools.playCoverContainer
+            .appendingPathComponent(info.displayName)
+            .appendingPathExtension("app")
```
amirsaam, line 190: Does this address #304 ?

IsaacMarovitz: Maybe, but it wasn't written to fix that bug

amirsaam: We will find out then 👍

amirsaam:
Shouldn't we have an `Add Source` button in the IPA Library View?
For better UX...

IsaacMarovitz:
Mm good point, I've got a couple of UX improvements I can make

amirsaam:
Since in-app download is implemented and we have IPA sources, does it check the IPA versions to notify newer versions available for update to user?

It can use notification function or an ⬆️ icon alongside apps that has been installed in IPA library

Amachik:
can we merge this?

IsaacMarovitz:
@amirsaam Currently it will take the most recent version in the IPA Source, it doesn't check the App Store currently

IsaacMarovitz:
@Amachik All of our reviewers are currently busy so merges have been slow

amirsaam:
> @amirsaam Currently it will take the most recent version in the IPA Source, it doesn't check the App Store currently

Why should it check App Store?
PlayCover should check App Library app version with the one from IPA Library that comes from IPA Source and tell there is difference in versions

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -109,31 +189,21 @@ struct StoreAppConditionalView: View {
             }
         }
         .task {
-            iconUrl = await getIconURLFromBundleIdentifier(app.id, app.region)
+            itunesData = await getITunesData(app.itunesLookup)
```
Depal1, line 192: Leaving the "itunesLookup" attribute empty `""` for apps not available in the App Store for instance, results in a crash. This isn't desired behavior, as PlayCover should let you list apps that are not available in the App Store.

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -0,0 +1,263 @@
...
+
+    func validateSource(_ source: String) {
+        sourceValidationState = .checking
+        DispatchQueue.global(qos: .userInteractive).async {
+            if let url = URL(string: source) {
+                newSourceURL = url
+                if StoreVM.checkAvaliability(url: newSourceURL!) {
+                    do {
+                        if newSourceURL!.scheme == nil {
+                            newSourceURL = URL(string: "https://" + newSourceURL!.absoluteString)!
```
Depal1, line 233: Local links are not being accepted as valid in the current scheme. `file:///path/to/my/apps/my_apps.json`

Depal1 (changes requested):
Having app entries in the JSON with wrong information (like a mistyped link in `"itunesLookup"`) throws an error as it should, but if there are several non compliant entries there will be as many issue prompts, which in practice does not provide any more useful information to the user. Additionally, once the view is changed to grid, list, or from App Library, all the prompts are shown again in the same session.

When no internet connection is active, the issue describing `The internet connection appears to be offline` does not get triggered if all entries in the JSON files are compliant. On the other hand, when there are `n` non compliant app entries in a JSON, this issue is prompted `n` times.

Depal1:
> > @amirsaam Currently it will take the most recent version in the IPA Source, it doesn't check the App Store currently
> 
> Why should it check App Store? PlayCover should check App Library app version with the one from IPA Library that comes from IPA Source and tell there is difference in versions

That is out of scope. PlayCover shouldn't monitor every app in a JSON list for updates from the App Store since it can't do anything meaningful about it (we can't fetch apps from the App Store), so it would amount to wasteful resource usage. Checking for updates is an exclusive task of the user.

amirsaam:
@Depal1 hmmm, why would it check app store?
`ipa version in json >  ipa version installed = update available`

Depal1:
> @Depal1 hmmm, why would it check app store?
> `ipa version in json > ipa version installed = update available`

Oh, okay. That is more sensible. Sure, it could be done.

PlayCover/Views/Settings/IPASourceSettings.swift
IsaacMarovitz: Do we really need to support local links?

Depal1 (approved):
Related features that are worth considering for future PRs:
- [ ] Local JSON support as local links.
- [x] Version number check against JSON file.

PR #370, opened 2022-09-22
Adding keymapping download

ZhichGaming:
Checks if the keymapping for a game exists in Puck's keymapping JSON, requires https://github.com/PlayCover/Puck/pull/10 to be merged.

PR #371, opened 2022-09-22
Adding keymapping download

ZhichGaming:
Checks if the keymapping for a game exists in the keymaps repository, #368 should be merged first. Repost of #370 since that one had unrelated history. 

- [x] Add keymaps to PlayCover/keymaps
- [x] Modify code to support github repositories instead of Puck json
- [x] Add custom sources (After #368 is merged)


amirsaam:
did you consider games having multiple keymaps in the json?

ZhichGaming:
No, I did not get to that yet. Thanks for making me aware of it though.

IsaacMarovitz:
I'm not sure this is a good idea. If we add this, we should allow the user to configure key mapping sources similar to how we're changing IPA Sources to work. Maybe it would be better to just add keymap files as an optional lines to IPA source jsons.

ZhichGaming:
I don't think there would be any sources that provide keymapping for PlayCover, as IPAs have a few sources like ipadecrypt.day or armconverter. We could add a keymapping folder in the PlayCover repo though? I don't understand why we would be not able to provide keymapping. Even if it's because it's not an important feature for us, it is a nice addition that would not really require any extra support.

PlayCover/en.lproj/Localizable.strings
```diff
@@ -86,6 +86,10 @@
 "settings.tab.km" = "Keymapping";
 "settings.toggle.km" = "Keymapping";
 "settings.toggle.km.help" = "Add keyboard support to touch-only games. Use Command + K to open editor menu.";
+"settings.button.km.download" = "Download keymapping";
+"settings.button.km.unavailable" = "No keymapping";
+"settings.button.km.loading" = "Loading...";
+"settings.button.km.forGame" = "Keymapping for";
```
amirsaam, line 140: I think "Keymap" is more accurate.
And the Unavailable string better to be: "Not yet provided"
for Loading: "Searching..."

ZhichGaming: I feel like "Keymapping" is fine as we were using it before. I also feel like it just sounds better in this context. We can change it through Weblate in the future if you're dissatisfied with it.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -29,14 +36,26 @@ class StoreVM: ObservableObject {
...
+            encode(source: keymappingSources)
+        }
+    }
 
     let sourcesUrl: URL
+    let keymappingSourcesUrl: URL
 
     @discardableResult
-    public func decode() -> Bool {
+    public func decode(source: [SourceData]? = nil) -> Bool {
```
IsaacMarovitz: I'm not sure about this approach. I think it might be better to just give them each separate functions.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -84,8 +109,6 @@ class StoreVM: ObservableObject {
     }
 
     func resolveSources() {
-        if !NetworkVM.isConnectedToNetwork() { return }
```
IsaacMarovitz: Don't remove this check

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -127,34 +150,68 @@ class StoreVM: ObservableObject {
         fetchApps()
     }
 
-    func deleteSource(_ selected: inout Set<UUID>) {
-        self.sources.removeAll(where: { selected.contains($0.id) })
+    func deleteSource(_ selected: inout Set<UUID>, source: [SourceData]? = nil) {
```
IsaacMarovitz: Same here, separate functions would be better, same goes for the rest of them.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -116,6 +196,108 @@ struct KeymappingView: View {
...
+        }
+        .onChange(of: showImportSuccess) { _ in
+            ToastVM.shared.showToast(
+                toastType: .notice,
+                toastDetails: NSLocalizedString("alert.kmImported", comment: ""))
+        }
+    }
+
+    // Check if the game selected is in keymaps repository
+    func isInKeymaps() async {
```
IsaacMarovitz, line 252: Add a network check to this function

IsaacMarovitz (changes requested):

PlayCover/ViewModel/StoreVM.swift
ZhichGaming: Ah, I might've accidentally deleted that during merge.

Issue #155, opened 2022-08-13
Clear app data is not comprehensive


Depal1:
Pressing the clear app data button seems to only clear "~/Library/Containers/BUNDLE_ID”, instead of clearing other data related to the app.

For instance:

```
~ % brew createzap plague        
zap trash: [
  "~/Library/Application Scripts/com.ndemiccreations.plagueinc",
  "~/Library/Caches/com.ndemiccreations.plagueinc",
  "~/Library/Containers/com.ndemiccreations.plagueinc",
  "~/Library/HTTPStorages/com.ndemiccreations.plagueinc",
]
```
Now pressing the clear app data button.
```
~ % brew createzap plague
zap trash: [
  "~/Library/Application Scripts/com.ndemiccreations.plagueinc",
  "~/Library/Caches/com.ndemiccreations.plagueinc",
  "~/Library/HTTPStorages/com.ndemiccreations.plagueinc",
]
```
Another directory that is created for some apps, and it isn’t deleted either is `"~/Library/Saved Application State/BUNDLE_ID.savedState"`

JoseMoreville:
I'm pretty sure this feature may be of help for this case https://github.com/PlayCover/PlayCover/pull/156

JoseMoreville:
Otherwise, clear app data might be better to be an implementation of https://github.com/PlayCover/PlayCover/pull/156 + removing all stored users + game assets.

Depal1:
This issue hasn’t been addressed yet.

PR #383, opened 2022-09-29
App uninstall settings

TheMoonThatRises:
Add settings for how Playcover uninstall apps. This includes removing keymapping, app settings, entitlements, and/or app containers. There is also an option to remove unused files and containers (unused as in the app is not installed).
Screenshot 2022-09-29 at 12 46 32


amirsaam:
Is it not better to have this options o the uninstall confirmation pop up?
This way users needs to open setting and change the setting everytime they want to delete an app but on the pop up they just need to check the options for that specific app.

TheMoonThatRises:
I could use both features, as the one in settings could as a default. The popup could be part of the confirm uninstall alert.

Depal1:
You might want to take a look at this too https://github.com/PlayCover/PlayCover/issues/155.

TheMoonThatRises:
> You might want to take a look at this too https://github.com/PlayCover/PlayCover/issues/155.

Do you want me to address it in this PR it a separate one?

TheMoonThatRises:
The uninstall popup currently looks like this.
uninstall_popup

amirsaam:
is there a way to undo the `Don't ask again` option in PlayCover?
I think it's not necessary to have that now with this new options.
It's better to not have it than to have it

btw the strings should be simpler than the PlayCover Preferences like this:
- Clear app data
- Remove keymappings
- Remove settings
- Remove entitlements

Depal1:
> > You might want to take a look at this too #155.
> 
> Do you want me to address it in this PR it a separate one?

If you want to address it, it could be better to implement it on a separate PR, unless it conflicts with your current implementation.

TheMoonThatRises:
> is there a way to undo the Don't ask again option in PlayCover?

Yes, there's a way to undo not displaying the uninstall alert in settings.

>  the strings should be simpler than the PlayCover Preferences like this:

Done.

amirsaam:
So if it undo it, can we have a note under it in small size (subtitle) to mention it?
Like `You can change this in the PlayCover Preferences` or smth like it.

TheMoonThatRises:
The suppression checkbox is an NSButton, so it can't have a subtitle. Would it be ok if the text was part of the title for the checkbox?

amirsaam:
I'm afraid that it will not follow the guidelines but can we have a hoverable/clickkable `U+1F6C8` (🛈) as part of the checkbox to show the message in a tooltip?

TheMoonThatRises:
Is this good?
uninstall_suppression_tooltip

amirsaam:
Uhh, I can't figure out how it would be triggered?
Is it a pointer on text hover? If yes, it's hover trigger time should be => 0.5s

TheMoonThatRises:
The NSButton (which is what the suppression checkbox is made from) has a tooltip option. You would trigger it by hovering over the text.

amirsaam:
I see. I think it will do the job. Thanks for being nice to do the extra work 👍

TheMoonThatRises:
Need to do more work on this PR since #398 is merged.

PlayCover/en.lproj/Localizable.strings
```diff
@@ -79,6 +88,11 @@
 "alert.legacyImport.subtitle" = "Legacy settings and keymapping files were detected. Would you like to convert these files to the new format?";
 "alert.wrongFileType.title" = "Wrong file type!";
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
+"alert.uninstall.clearAppData" = "Clear app data";
+"alert.uninstall.removeKeymap" = "Remove app keymapping";
```
```suggestion
"alert.uninstall.removeKeymap" = "Remove app keymap";
```

PlayCover/en.lproj/Localizable.strings
```diff
@@ -23,6 +23,15 @@
 "preferences.button.checkForUpdates" = "Check for updates now...";
 "preferences.tab.updates" = "Updates";
 "preferences.toggle.automaticUpdates" = "Automatically check for updates";
+"preferences.tab.uninstall" = "Uninstall";
+"preferences.toggle.showUninstall" = "Show uninstall popup";
+"preferences.toggle.clearAppData" = "Clear app data when uninstalling";
+"preferences.toggle.removeKeymap" = "Remove app keymapping when uninstalling";
```
```suggestion
"preferences.toggle.removeKeymap" = "Remove app keymap when uninstalling";
```

PlayCover/Views/Settings/UninstallSettings.swift
```diff
@@ -0,0 +1,53 @@
...
+
+struct UninstallSettings: View {
+    public static var shared = UninstallSettings()
+
+    @ObservedObject var uninstallPreferences = UninstallPreferences.shared
+
+    @State private var showPruneFileAlert = false
+
+    var body: some View {
+        Form {
```
Depal1, line 29: Having "when uninstalling" on each line is redundant. It would be better to have something like this:
Preview settings
Saying perhaps `When uninstalling an app:`

PlayCover/en.lproj/Localizable.strings
```diff
@@ -79,6 +88,11 @@
 "alert.legacyImport.subtitle" = "Legacy settings and keymapping files were detected. Would you like to convert these files to the new format?";
 "alert.wrongFileType.title" = "Wrong file type!";
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
+"alert.uninstall.clearAppData" = "Clear app data";
```
Depal1: I think we can remove `app` from these strings since it is derived from context.

Depal1 (changes requested):
Featurewise this is working as intended. Prune is a nice touch.

Depal1 (approved):

PR #387, opened 2022-09-30
Force RTL while Language is RTL

amirsaam:
**This is an UX improvement PR**

Choosing a RTL language as Preferred Language that macOS itself is not translated to will result in an app with translated language but not a correct `LayoutDirection`.
It is not a good experience for user to be forced to change the app specific language to have the correct layout.

_Only known RTL language that macOS has translation is Arabic, never test apps under this language for RTL compatibility._ If you wanna test, test it with Persian.

Since you are forcing the app to become accept RTL, subviews are not accepting that and so we need to have a custom `.environment` value in those views:
```swift
.environment(\.layoutDirection, direction)
```
`direction` is a public variable that has `LayoutDirection` as value that is either RTL or LTR based on preferred language.
And we need to use `.flipsForRightToLeftLayoutDirection(true)` for some UI element like mouse sensitivity slider to be flipped only when in use language is RTL.
You can check direction as Boolean with public function `isRTL()` if you need it (`rtl == true`).

Tasks:
- [x] Having an easy to setup property for future UI developments
- [ ] Applying new property to all current views
- [ ] Checking for broken UI elements that needs to be flipped for RTL
- [x] Adding both RTL and LTR to Xcode Preview`*`

Current situation:
- Toolbar is out of position and has incorrect direction
- `AppSetting` view's `TabView` is not ordered from Right to Left
- Genshin Impact account manager views need to be tested (I don't have it)`*`
- `ToastView` for XCLT prompt is not tested yet`*`
- Menubar items not RTL (very low priority, maybe never fix it because its language is mixed)

Screenshot_1401-07-07_at_04 58 24

`*` Xcode Previews have a problem with `Sparkle`, error:
```swift
HumanReadableSwiftError

SettingsError: noExecutablePath(<IDESwiftPackageStaticLibraryProductBuildable:ObjectIdentifier(0x00000002cb9b87f0):'Sparkle'>)
```
And because of it I cannot test Genshin Account Manager or XCLT installer.


PlayCover/Views/PlayCoverApp.swift
```diff
@@ -31,6 +31,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
 }
 
+func dirCheck() -> Bool {
+    let langCode  = Locale.current.languageCode
+    let rtlLangs = [ "ar", "arc", "dv", "fa", "ha", "he", "khw", "ks", "ku", "ps", "ur", "yi" ]
+    if rtlLangs.contains(langCode!) { return true } else { return false }
```
JoseMoreville: You can simplify this line by returning boolean computation

JoseMoreville (changes requested):

PlayCover/Views/PlayCoverApp.swift
amirsaam: done.

amirsaam:
Ok, updated the PR information with details.
@Depal1 can you please add `help-wanted` label to this pr?

zanderp25:
> can you please add `help-wanted` label to this pr?

I gotchu

zanderp25:
Okay, so i tested a few things:

> Genshin Impact account manager views need to be tested (I don't have it) `*`

I built your PR on my computer after setting the language to Arabic and I tested the account views on Genshin

‏لقطة الشاشة ٢٠٢٢-١٠-٠٩ في ١٤ ٢٧ ٣٥
‏لقطة الشاشة ٢٠٢٢-١٠-٠٩ في ١٤ ٣٣ ١٦

It looks fine to me

> AppSetting view's TabView is not ordered from Right to Left

Looks fine to me

‏لقطة الشاشة ٢٠٢٢-١٠-٠٩ في ١٤ ٢٨ ٤٢

> Menubar items not RTL (very low priority, maybe never fix it because its language is mixed)

This also looks fine to me

‏لقطة الشاشة ٢٠٢٢-١٠-٠٩ في ١٤ ٤١ ٤٥

amirsaam:
Thanks @zanderp25 but like I said:

> Only known RTL language that macOS has translation is Arabic, never test apps under this language for RTL compatibility.

I appreciate what you have done.

zanderp25:
> Only known RTL language that macOS has translation is Arabic, never test apps under this language for RTL compatibility.

Why not? I didn't reboot my computer so it only changes the language of programs that are relaunched after the change.

Edit: Also, I do know that macOS supports Hebrew (another RTL lang), but PlayCover does not.

amirsaam:
> Why not? I didn't reboot my computer so it only changes the language of programs that are relaunched after the change.

Well, because macOS itself has been translated to Arabic and Hebrew (I didn't know about this one) the whole system becomes rtl and cuz of that you'll see no problem on ui, but other languages that have no translation for macOS will have a lot of problems; test it under Persian, you'll see.

PR #398, opened 2022-10-04
More extensive clear app data

TheMoonThatRises:
Solves #155. Needs to be merged before #383. Are there any other files that should be removed with clear app data (like should the keymappings, app settings, etc be included)?


amirsaam:
Since your full featured uninstall is there, I don't think it would be needed to clear settings and etc. because if someone wanted to clear all of those, then he really needs to reinstall instead of clearing app data

Depal1 (approved):

PR #413, opened 2022-10-10
Install apps without injecting PlayTools

TheMoonThatRises:
Allows the user to install apps without always injecting PlayTools. This might take a while to complete.

- [x] Allow user to distinguish between apps with and without PlayTools
- [x] Someway for the code to differentiate between PlayToolsless apps
- [x] Allow users to quickly inject and remove PlayTools from apps


IsaacMarovitz:
Could you add either a toggle/button in either App Settings or the context menu to allow the user to easily opt in or out of PlayTools on the fly without reinstalling? 

> Someway for the code to differentiate between PlayToolsless apps

I would suggest adding a `isPlayToolsInstalled` Bool to the App Settings plist

Depal1:
As for a way for the user to distinguish whether an app is PlayTools less or not, a bubble with an appropriate icon could be placed in the corner of the app icon, plus an entry for ´isPlayToolsInstalled´ in the info tab.

TheMoonThatRises:
> Could you add either a toggle/button in either App Settings or the context menu to allow the user to easily opt in or out of PlayTools on the fly without reinstalling?

How would PlayTools be removed after it has already been in the executable?

TheMoonThatRises:
> I would suggest adding a `isPlayToolsInstalled` Bool to the App Settings plist

Also the one thing I worry about adding another thing to the App Settings is that it would reset the user's current settings, which may not be fun for the user.

IsaacMarovitz:
> > Could you add either a toggle/button in either App Settings or the context menu to allow the user to easily opt in or out of PlayTools on the fly without reinstalling?
> 
> How would PlayTools be removed after it has already been in the executable?

You would have to use `optool` to remove the added load command from the dylib, and also remove the `PlugIns` folder from the .app (This is mostly optional but if you want to return the bundle to its original state, it's necessary)

IsaacMarovitz:
> Also the one thing I worry about adding another thing to the App Settings is that it would reset the user's current settings, which may not be fun for the user.

On further thought, App Settings are also designed to be transferable from user to user so people can easily share configurations. In that case, it might be worth just creating a plist for PlayTools install status

PlayCover/AppInstaller/Installer.swift
```diff
@@ -8,8 +8,35 @@
...
+        let yes = alert.addButton(withTitle: NSLocalizedString("button.Yes", comment: ""))
+        alert.addButton(withTitle: NSLocalizedString("button.No", comment: ""))
+
+        yes.hasDestructiveAction = true
+
+        let response = alert.runModal()
+
+        if alert.suppressionButton?.state == .on {
+            InstallSettings.shared.showInstallPlayToolsPopup = false
+            InstallSettings.shared.alwaysInstallPlayTools = response == .alertFirstButtonReturn
```
IsaacMarovitz, line 29: This option feels largely redundant. If you always install PlayTools, why ask the user whether or not to install it? It feels like the show PlayTools popup shouldn't appear by default, and shouldn't be available if this option is set to true.

PlayCover/Views/Settings/PlayCoverSettingsView.swift
```diff
@@ -13,7 +13,7 @@ struct PlayCoverSettingsView: View {
     @EnvironmentObject var storeVM: StoreVM
 
     private enum Tabs: Hashable {
-        case updates, ipasource, uninstall
+        case updates, ipasource, uninstall, install
```
IsaacMarovitz: On my machine, the layout appears in the order Updates, Uninstall, IPA Sources, Install. Not sure if it's just a me thing or not.
Screenshot 2022-10-19 at 22 36 36

PlayCover/Views/App Views/PlayAppView.swift
```diff
@@ -207,6 +207,13 @@ struct PlayAppConditionalView: View {
                     Text(app.name)
                         .foregroundColor(selected?.info.bundleIdentifier == app.info.bundleIdentifier ?
                                          selectedTextColor : Color.primary)
+
```
IsaacMarovitz, line 216: The style of this exclamation point doesn't match up with the badge in IPA Library. In IPA Library, the icon is unfilled, whereas here, it is filled. In IPA Library, the icon is on the left, here, it's on the right. In IPA Library, it's included in the selection, here, it is not.

Screenshot 2022-10-19 at 22 39 30
Screenshot 2022-10-19 at 22 39 36

IsaacMarovitz (changes requested):

IsaacMarovitz:
Screenshot 2022-10-19 at 22 40 45

I'm not entirely convinced by the placement of the warning in App Settings, maybe it would work better if it was right-aligned, and in line with the title.

PlayCover/Views/Settings/PlayCoverSettingsView.swift
IsaacMarovitz: Updates, IPA Sources, Install, Uninstall makes more sense to me

Depal1:
> Screenshot 2022-10-19 at 22 40 45
> 
> I'm not entirely convinced by the placement of the warning in App Settings, maybe it would work better if it was right-aligned, and in line with the title.

Wouldn't that be conflictive in apps with large names? I just state it as a possibility. I think it is okay as is.

PlayCover/AppInstaller/Installer.swift
TheMoonThatRises: The `alwaysInstallPlayTools` is only used when the user suppresses PlayTools install popup. When it is suppressed, it installs PlayTools based on the option they chose when they suppressed the popup. When the popup is used, it installs PlayTools based on what the user choses.

IsaacMarovitz:
> Wouldn't that be conflictive in apps with large names? I just state it as a possibility. I think it is okay as is.

I think mainly the spacing just looks a little off. Maybe something like this would work better?


196843562-c000d579-a162-48f0-90ae-2c2c4292ded9

Depal1:
> > 
> I think mainly the spacing just looks a little off. Maybe something like this would work better?
> 
> 196843562-c000d579-a162-48f0-90ae-2c2c4292ded9

I think I noticed the spacing issue you mention. Aligning the warning just under the first letter of the app name could be an option, instead of under the app icon. Putting the warning in the lower end wouldn't be noticeable.

TheMoonThatRises:
> I think I noticed the spacing issue you mention. Aligning the warning just under the first letter of the app name could be an option, instead of under the app icon. Putting the warning in the lower end wouldn't be noticeable.

Is this spacing better?
Screenshot 2022-10-20 at 08 43 05

Depal1:
> >
> Is this spacing better?

It is. @IsaacMarovitz What do you think?

amirsaam:
text should've less spacing from the warning sign i would say. like:
`⚠️.PlayTools`
now it's like
`⚠️...PlayTools`

PlayCover/AppInstaller/Installer.swift
IsaacMarovitz: Maybe the `alwaysInstallPlaytools` option should be made disabled when `showInstallPlaytoolsPopup` is enabled to make its functionality clearer.

PlayCover/Views/App Views/PlayAppView.swift
```diff
@@ -181,14 +180,26 @@ struct PlayAppConditionalView: View {
...
     @State var iconURL: URL?
-    @State var app: PlayApp
     @State var isList: Bool
 
     var body: some View {
         Group {
             if isList {
                 HStack(alignment: .center, spacing: 0) {
+                    let noPlayToolsImage = Image(systemName: "exclamationmark.triangle.fill")
```
IsaacMarovitz: This layout works well for IPA Library, but when the icon only shows up occasionally, the extra spacing looks a little off.

IsaacMarovitz: Either make these icons `exclamationmark.triangle` or make the IPA Library download icon `arrow.down.circle.fill`

IsaacMarovitz (changes requested):

PlayCover/Views/App Views/PlayAppView.swift
TheMoonThatRises: Do you mean this extra spacing?
Screenshot 2022-10-20 at 17 32 53

IsaacMarovitz: Yeah

TheMoonThatRises: Would it just be better to have it offset? (Also should the padding be smaller between the icon and the app image?)
Screenshot 2022-10-20 at 19 36 11

amirsaam: is it not better to change the location of the sign to the trailing side of the game/app icon?

IsaacMarovitz: @TheMoonThatRises I think just offsetting is good enough for now

IsaacMarovitz: Make sure the padding is consistent tho

Depal1 (approved):

PR #426, opened 2022-10-18
Missed translation

zanderp25:
I missed a single string while translating the app to Spanish and cannot submit through Weblate because it is locked for some reason.

Edit: I added the "no sources" UI strings because I can't use Weblate.


IsaacMarovitz:
Weblate should be open again

PR #417, opened 2022-10-12
Code Cleanup + Styling Fixes + Image Cache

IsaacMarovitz:
This PR fixes #362 
This also fixes issues where Duolingo, Infinity Blade, and Netflix would show with the wrong/no icon

PR #429, opened 2022-10-19
SourceVM Backend

IsaacMarovitz:
A new implementation of #371, based on the #417 branch

Reasons for reimplementing:
- Zhich's implementation suffers from a couple of structural problems that go counter to the existing source-based VM infrastructure
- I initially attempted to add my fixes in a PR to Zhich's branch but I kept running into weird state issues, so in the end I felt that reimplementing would be easier, and reduce spaghetti


Depal1:
@IsaacMarovitz @ZhichGaming I don't know at what extent will the code of #371 be used as a reference, but in any case I will leave this here.
https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors
https://gist.github.com/lisawolderiksen/f9747a3ae1e58e9daa7d176ab98f1bad

PR #450, opened 2022-10-26
TikTok Blacklisting and uninstall function.

Amachik:
simple solution. should work ok.


Candygoblen123:
Can you fix the swiftLint errors?

PlayCover/en.lproj/Localizable.strings
```diff
@@ -113,6 +113,7 @@
 "error.appEncrypted" = "This app is encrypted! Please use a decrypted IPA. iMazing IPAs are not supported!";
 "error.appCorrupted" = "Something went wrong with this IPA. Please try to use another IPA file.";
 "error.appProhibited" = "Using this app through PlayCover will likely result in a ban!";
+"error.appProhibitedTiktok" = "TikTok behaves maliciously on Macs & using it can leave your device in an unbootable state, therefore is blocked from PlayCover. (Also.. Uninstalling this devil's app...)";
```
```suggestion
"error.appProhibitedTiktok" = "TikTok behaves maliciously on Macs, and using it can leave your device in an unbootable state, therefore it is blocked from PlayCover.";
```
Although, a more general message would be better.

Depal1 (changes requested):

PlayCover/en.lproj/Localizable.strings
Depal1: The references should be general, and not specifically referenced as TikTok.

Depal1:
This PR is reverting changes from https://github.com/PlayCover/PlayCover/pull/449. Please rebase.

PlayCover/Model/PlayApp.swift
```diff
@@ -162,6 +167,9 @@ class PlayApp: BaseApp, ObservableObject {
         "com.pubg.newstate",
         "com.tencent.tmgp.pubgmhd",
         "com.dts.freefireth",
-        "com.dts.freefiremax"
-    ]
+        "com.dts.freefiremax",
+]
+    static let PROHIBITED_APPS_TIKTOK = [
```
TheMoonThatRises: Riding on Depal's comment, this should be generalised to `MALICIOUS_APPS` (or something similar) so that the list can be expanded as new apps are discovered.

PlayCover/en.lproj/Localizable.strings
Amachik: we should also inform the user about the uninstallation

Depal1: A rephrase is needed then. In order to make this feature future proof, it must be a general message that could work for more apps.

Amachik: is in it better to just rewrite PROHIBITED_APPS to do the uninstall and message? would solve every issue. instead of creating one specific for tiktok

PlayCover/Model/PlayApp.swift
```diff
@@ -112,6 +116,7 @@ class PlayApp: BaseApp, ObservableObject {
 
     func clearAllCache() {
         Uninstaller.clearExternalCache(info.bundleIdentifier)
+        Uninstaller.clearExternalCache(info.bundleName)
```
TheMoonThatRises: This line does not need to be here. The name of the folder of all the caches are always the bundle identifier.

Amachik: ty for info. 💯

PlayCover/Model/PlayApp.swift
TheMoonThatRises: Also, to keep consistency with `PROHIBITED_APPS`, it would be better to use the bundle identifier instead of the name of the app.

PlayCover/en.lproj/Localizable.strings
Depal1: Uninstall should be reserved for malicious apps only. Hence @TheMoonThatRises's suggestion would be better.

ZhichGaming:
It would be better if you made an alert pop up that only has the option to uninstall instead of just uninstalling as soon as it is launched. The user would be able to either force quit the app or proceed with the uninstall. This way, the user would have less risk of losing data and they wouldn’t panic as much.

Amachik:
i will do a rewrite of this.

> 

moons solution is the best i would say

Amachik:
> It would be better if you made an alert pop up that only has the option to uninstall instead of just uninstalling as soon as it is launched. The user would be able to either force quit the app or proceed with the uninstall. This way, the user would have less risk of losing data and they wouldn’t panic as much.

also there will be no data cuz the app will not even open.

TheMoonThatRises:
> also there will be no data cuz the app will not even open.

This is more for users who upgraded and have already used the app before it was tagged as malicious.

PlayCover/Model/PlayApp.swift
Amachik: thats optional i would say. i can do that for sure. in my opinion, bundle.name is better than bundle.identifier.

ZhichGaming:
Ah, that’s true I guess. But maybe the user wants to bypass it, ignoring the warning by opening through finder. We wouldn’t want them accidentally deleting their data if they opened it accidentally through playcover.

Amachik:
> Ah, that’s true I guess. But maybe the user wants to bypass it, ignoring the warning by opening through finder. We wouldn’t want them accidentally deleting their data if they opened it accidentally through playcover.

if user is messing with files bcs he wants to bypass it, thats not really ur problem. he is on his own. he can use sideloadly.

Amachik:
All 
should be fixed.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -78,14 +76,14 @@ struct StoreAppView: View {
                     .appendingPathExtension("ipa")
                 try FileManager.default.moveItem(at: url, to: tmpDir)
                 uif.ipaUrl = tmpDir
-                DispatchQueue.main.async {
```
TheMoonThatRises, line 81: This is still reverting the changes done in #449.

Amachik: wtf i did clean fork :D i feel like an idiot rn .

Amachik:
i need to close this. no other way to fix this issue.

PR #453, opened 2022-10-26
Improve PlayTools installer interactions

TheMoonThatRises:
Makes the PlayTools install popup clearer by including what it does, and not appear to the regular user

* Make install popup not popup by default, and triggered by holding option key
* Clarify what PlayTools does and not make the button destructive
* Allow user to cancel installation process when using the install popup
* Remove AKInterface if PlayTools is not installed to restore app to it's original state

- [x] Make install popup appear only when the user presses the option key while installing or configures it in settings
- [x] Improved install alert

### DIFF
--- a/PlayCover/en.lproj/Localizable.strings
+++ b/PlayCover/en.lproj/Localizable.strings
@@ -95,7 +95,7 @@
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
 "alert.install.injectPlayTools" = "Inject PlayTools";
 "alert.install.injectPlayToolsQuestion" = "Do you want to inject PlayTools into this application?";
-"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but may break certain apps";
+"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but it may break certain apps";
 "alert.supression" = "You can change this in the PlayCover preferences";
 
 "button.OK" = "OK";
--- a/PlayCover/en.lproj/Localizable.strings
+++ b/PlayCover/en.lproj/Localizable.strings
@@ -94,7 +94,7 @@
 "alert.wrongFileType.title" = "Wrong file type!";
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
 "alert.install.injectPlayTools" = "Do you want to inject PlayTools into this application?";
-"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more";
+"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but may break certain apps";
 "alert.supression" = "You can change this in the PlayCover preferences";
 
 "button.OK" = "OK";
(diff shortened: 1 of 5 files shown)


Hulkstern (approved):
The changes to the playtools prompt are definitely needed clarification on the prompt.

(This was meant to be a comment, don't know why it went through as me "approving" these changes since I don't have write access)

PlayCover/en.lproj/Localizable.strings
```diff
@@ -93,8 +93,8 @@
 "alert.legacyImport.subtitle" = "Legacy settings and keymapping files were detected. Would you like to convert these files to the new format?";
 "alert.wrongFileType.title" = "Wrong file type!";
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
-"alert.install.injectPlayTools" = "Inject PlayTools";
-"alert.install.injectPlayToolsInApp"= "Do you want to inject PlayTools into this application?";
+"alert.install.injectPlayTools" = "Do you want to inject PlayTools into this application?";
+"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more";
```
Candygoblen123: Can you add something like "However, it may break some apps."

Candygoblen123 (dismissed):

PlayCover/en.lproj/Localizable.strings
TheMoonThatRises: Is it better now?

TheMoonThatRises:
Is an install popup like this good, or would it not be needed as there won't be enough install options for this to be useful?

Screenshot 2022-10-30 at 20 58 41

Depal1:
> Is an install popup like this good, or would it not be needed as there won't be enough install options for this to be useful?
> 
> Screenshot 2022-10-30 at 20 58 41

At present, I don't think there are other install options we are considering.

TheMoonThatRises:
> At present, I don't think there are other install options we are considering.

So should I keep the other popup?

Candygoblen123:
> Is an install popup like this good, or would it not be needed as there won't be enough install options for this to be useful?
> 
> Screenshot 2022-10-30 at 20 58 41

Can you change the tooltip to be just text in the popup?

TheMoonThatRises:
> Can you change the tooltip to be just text in the popup?

Like this? The main reason I have it as a tooltip is because it wouldn't be necessary text to most people who would use the menu, as they would already know what PlayTools is and does.
Screenshot 2022-10-31 at 13 26 27

amirsaam:
I think what candy meant is like uninstall pop up that asks if you are sure of deleting the app text. but since playtools is the only install option as @Depal1 said, it will be good if it only asks about PlayTools like:

Headline:
**Inject Playtools?**

Subheadline:
**_The Description_**

Buttons:
1. Inject & Install
2. Install w/o Playtools
3. Cancel

TheMoonThatRises:
So something like this?
Screenshot 2022-10-31 at 18 47 11

amirsaam:
It's better now but I think the options I suggested were more clear to user comparing to simple yes or no buttons, and could simplify the title.
But it's up to maintainers to decide. Thanks for your effort 👍

PlayCover/en.lproj/Localizable.strings
```diff
@@ -94,7 +94,8 @@
 "alert.wrongFileType.title" = "Wrong file type!";
 "alert.wrongFileType.subtitle" = "PlayCover can only install valid .ipa files.";
 "alert.install.injectPlayTools" = "Inject PlayTools";
-"alert.install.injectPlayToolsInApp"= "Do you want to inject PlayTools into this application?";
+"alert.install.injectPlayToolsQuestion" = "Do you want to inject PlayTools into this application?";
+"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but may break certain apps";
```
```suggestion
"alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but it may break certain apps";
```

Depal1 (approved):

PR #475, opened 2022-11-01
repull of #450.

Amachik:
problem with merging via desynced files.


Amachik:
it should be all fixed now. i hope.

Amachik (commented):
looks ok to me.

TheMoonThatRises:
> it should be all fixed now. i hope.

Unfortunately, there is one in StoreAppView that still isn't fixed.

Amachik:
> > it should be all fixed now. i hope.
> 
> Unfortunately, there is one in StoreAppView that still isn't fixed.

huh? what is it this time? im not discarding any changes

Amachik (commented):
im done with it >}

amirsaam:
Can we also prohibit users from installing it?
Malicious apps are different with regular prohibited, those are ban risks and users can bypass it with no real damage, but malicious are red flag, users should not be able to install it through PlayCover.

Amachik:
> Can we also prohibit users from installing it?
> Malicious apps are different with regular prohibited, those are ban risks and users can bypass it with no real damage, but malicious are red flag, users should not be able to install it through PlayCover.

i can def. add this feature in addition to this. i will add this later on. this pr is now complete.

Depal1 (approved):

PR #484, opened 2022-11-02
Fix PlayTools Installer

IsaacMarovitz:


PlayCover/Utils/PlayTools.swift
```diff
@@ -171,15 +171,7 @@ class PlayTools {
...
-            let bundleTarget = pluginUrl
-                .appendingPathComponent("AKInterface")
-                .appendingPathExtension("bundle")
-
-            if FileManager.default.fileExists(atPath: bundleTarget.path) {
-                try FileManager.default.removeItem(at: bundleTarget)
-            }
-
-            if try FileManager.default.contentsOfDirectory(atPath: pluginUrl.esc).isEmpty {
+            if FileManager.default.fileExists(atPath: pluginUrl.path) {
```
TheMoonThatRises, line 176: The PlugIn directory shouldn't be directly removed, since some apps use it.

TheMoonThatRises: But then again, some apps don't have the directory, so it would need to be deleted if it was created by PlayCover.

JoseMoreville (approved):

PR #494, opened 2022-11-04
Fancy IPA Library

IsaacMarovitz:
Controversial but some people like it, so I'm going to leave it here for consideration.

**Note: This requires macOS 13+**


PlayCover/Views/Sidebar Views/IPALibraryView.swift
```diff
@@ -20,98 +20,107 @@ struct IPALibraryView: View {
...
-                if !isList {
-                    LazyVGrid(columns: gridLayout, alignment: .center) {
-                        ForEach(storeVM.filteredApps, id: \.bundleID) { app in
-                            StoreAppView(selectedBackgroundColor: $selectedBackgroundColor,
-                                         selectedTextColor: $selectedTextColor,
-                                         selected: $selected,
-                                         app: app,
-                                         isList: isList)
-                            .environmentObject(DownloadVM.shared)
+        if #available(macOS 13.0, *) {
```
amirsaam, line 23: if you are using if available, why you need to bump macs requirement? on other macos version you can just simply start download.
I'm just asking

IsaacMarovitz: For some reason, SwiftUI requires the View to be available on macOS 12

amirsaam: then if you bumped requirements, why did you use if available? 🥴

PR #501, opened 2022-11-05
Reposition noPlayTools sign in ListView App Library

amirsaam:
Repositions the warn sign location in App Library's ListView for apps that have no PlayTools
Screenshot 1401-08-15 at 00 14 16


TheMoonThatRises:
The download icon positioning for the IPA library should be changed as well for consistency.

amirsaam:
> The download icon positioning for the IPA library should be changed as well for consistency.

Since they all do have one, I don't think changing it make the UI better but more chaotic.

IsaacMarovitz (approved):

Depal1 (approved):

PR #512, opened 2022-11-13
Release Temporary Directories

TheMoonThatRises:
Always delete temporary directories when PlayCover is finished using them. When installing an app fails, it leaves behind a directory with the IPA.

### DIFF
--- a/PlayCover/AppInstaller/Installer.swift
+++ b/PlayCover/AppInstaller/Installer.swift
@@ -50,9 +50,12 @@ class Installer {
         InstallVM.shared.next(.begin, 0.0, 0.0)
 
         DispatchQueue.global(qos: .userInitiated).async {
+            let ipa = IPA(url: ipaUrl)
+
             do {
-                let ipa = IPA(url: ipaUrl)
                 InstallVM.shared.next(.unzip, 0.0, 0.5)
+                try ipa.allocateTempDir()
+
                 let app = try ipa.unzip()
                 InstallVM.shared.next(.library, 0.5, 0.55)
                 try saveEntitlements(app)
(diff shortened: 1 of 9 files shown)


IsaacMarovitz:
Files and directories created in `.itemReplacementDirectory` will be automatically cleaned up by macOS

ohaiibuzzle:
Items under /var/folders are automatically cleaned up by macOS, at minimum, every reboots

Unless you install 25 apps in a row, 4 GB each, and keep trying to install despite them failing, it is not a huge concern

amirsaam:
For the records, there are people who don't reboot their device regularly.
Better to be cleaned, it's the professional way to do it.
Unless you wanna become Instagram and eat the storage with cache files on users phones...

p.s.: some people own only 128gb ssd

IsaacMarovitz:
> For the records, there are people who don't reboot their device regularly.
> 
> Better to be cleaned, it's the professional way to do it.
> 
> Unless you wanna become Instagram and eat the storage with cache files on users phones...
> 
> 
> 
> p.s.: some people own only 128gb ssd

Rebooting regularly isn't necessary either.

> Long answer:

> 1. launchd sets up /usr/sbin/periodic daily to run every 86400 seconds (1 day) (definition in /System/Library/LaunchDaemons/com.apple.periodic-daily.plist). Older versions of Mac OS scheduled it to run at 3:15 am each day.
> 2. /usr/sbin/periodic daily runs the scripts in /etc/periodic/daily, including 110.clean-tmps.
110.clean-tmps uses find to delete files not accessed (using the -atime and -mtime parameters to find) in $daily_clean_tmps_days days which aren't listed in $daily_clean_tmps_ignore.
Those two variables (and other related variables) are defined in /etc/periodic.conf. (Since OS 10.7 there is a file /etc/defaults/periodic.conf which defines the defaults, but for overwriting these defaults you should still create /etc/periodic.conf – see the periodic.conf(5) manual page.) If you haven't modified this file, $daily_clean_tmps_days is set to 3.
periodic daily logs its output to /var/log/daily.out. Files deleted from /tmp are logged under the heading "Removing old temporary files:".

TheMoonThatRises:
For my computer, the temporary directory is not cleaned by the daily cleaner, and only mentions cleaning /var/rwho (besides, PlayCover temporary files are stored under $TMPDIR, which is separate from /tmp). Besides, its better for PlayCover to clean itself up than to rely on the system or the user to do it.

Depal1 (approved):

PR #517, opened 2022-11-16
Remove keymap checks

ZhichGaming:
Removes forcing the user to rename their keymap and instead gives them the option to proceed with the import. 
Screen Shot 2022-11-16 at 2 12 40 AM


PlayCover/Utils/AlertExtensions.swift
```diff
@@ -0,0 +1,22 @@
...
+//  AlertExtensions.swift
+//  PlayCover
+//
+//  Created by Nick on 2022-11-16.
+//
+
+import AppKit
+
+extension NSAlert {
+    static func differentBundleIdKeymap(response: (NSApplication.ModalResponse) -> Void) {
```
TheMoonThatRises: Personally, I think this function should be in `Keymapping.swift` since it is very specific to keymapping. Also, it might be better to have the function return either `true`/`false` (depending on which button was pressed) instead of relying on it's response function. There can be a variable that then stores the value and handles it appropriately (which will also reduce complexity in the `importKeymap` function)

PlayCover/en.lproj/Localizable.strings
```diff
@@ -97,6 +97,8 @@
 "alert.install.injectPlayToolsQuestion" = "Do you want to inject PlayTools into this application?";
 "alert.install.playToolsInformative" = "PlayTools is a tool that allows you to bind keys to screen touches, custom display resolutions, and more, but it may break certain apps";
 "alert.supression" = "You can change this in the PlayCover preferences";
+"alert.differentBundleIdKeymap.message" = "The name of the keymap is different from the bundle ID of the app!";
```
TheMoonThatRises: I don't think most people will understand what bundle IDs are, so it would be better if `alert.differentBundleIdKeymap.text` was `alert.differentBundleIdKeymap.message` and have `alert.differentBundleIdKeymap.text` be something along the lines of "Are you sure you want to import this keymap?" so that the `Proceed` button makes more sense.

PlayCover/Utils/AlertExtensions.swift
ZhichGaming: Oh yeah you're right, that would be better than the current implementation. It would also remove some repetition.

PlayCover/Model/Keymapping.swift
```diff
@@ -9,6 +9,27 @@ import AppKit
 import Foundation
 import UniformTypeIdentifiers
 
+extension NSAlert {
+    static func differentBundleIdKeymap() -> Bool {
```
TheMoonThatRises: To make it more consistent with other alerts, it should be moved to be a private function in the Keymapping class instead of having it as an extension of NSAlert.

Depal1 (approved):

PR #527, opened 2022-11-20
Low Power Mode Alert

IsaacMarovitz:


Depal1:
If this is to be done, it should only show for specific models since always plugged models don’t have a low power mode.

Depal1:
At present, I don’t know how would it behave on the aforementioned Macs.

FlyMeToTheMoonAndLetMePlayAmongTheStars:
> At present, I don’t know how would it behave on the aforementioned Macs.

Would it just open normally if there is no low power mode support?

IsaacMarovitz:
> At present, I don’t know how would it behave on the aforementioned Macs.

For Macs that don't have a Low Power Mode, I imagine the bool is just always set to false.

KhoraLee:
> At present, I don’t know how would it behave on the aforementioned Macs.

It's old and iOS documentation but it is still linked in latest apple's developer documentation page. So it should be applied to mac too.
https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/LowPowerMode.html#//apple_ref/doc/uid/TP40015243-CH31
```
If the device’s power state is unknown or if the device doesn’t support Low Power Mode, then this property always has a value of NO.
```

Depal1 (approved):

PR #534, opened 2022-11-22
Source URL Schema Add

IsaacMarovitz:
Kinda sketchy but it works

### DIFF
--- a/PlayCover/Views/MainView.swift
+++ b/PlayCover/Views/MainView.swift
@@ -3,6 +3,8 @@
 //  PlayCover
 //
 
+// swiftlint:disable type_body_length
+
 import SwiftUI
 
 enum XcodeInstallStatus {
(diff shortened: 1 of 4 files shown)


Depal1 (commented):
URL schema should have some sort of manual deactivation/activation to avoid potential cases where it could pose as flag for apps as `running through PlayCover`.

ohaiibuzzle:
> URL schema should have some sort of manual deactivation/activation to avoid potential cases where it could pose as flag for apps as `running through PlayCover`.

This is near-impossible to do considering that every app has a handle registered to them. We already have one in the release, just that we don't notice it (`playcover:` will open PC)

The only way to do it as you says is to rely on collateral damage (aka register PlayCover as `tiktok:` or `facebook:` in hope that attempting detection will cause massive hassles)

PR #538, opened 2022-11-24
feat: (slightly) more robust uri handler

ohaiibuzzle:
This PR proposes a new, (slightly) more robust URI handler for the `playcover:` URI scheme, with support for adding sources directly from other applications

Also includes RuntimeVM, which is like StoreVM, but is not persistent, which can be useful for passing around temp data later.

Example: `playcover://source?action=add&url=<source_url>`

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "PlayCover/PlayTools" "3.0.0-staging"
+github "PlayCover/PlayTools" "51ab649983e977fd378706e0037866851ba538b8"
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "PlayCover/PlayTools" "51ab649983e977fd378706e0037866851ba538b8"
+github "PlayCover/PlayTools" "3.0.0-staging"
--- a/PlayCover/Utils/URLHandler.swift
+++ b/PlayCover/Utils/URLHandler.swift
@@ -38,7 +38,7 @@ struct URLHandler {
             let params = urlComponenents.queryItems else {
                 // Fall back to old url handler (for files)
                 if url.pathExtension == "ipa" {
-                    Installer.install(ipaUrl: uif.ipaUrl!, export: false, returnCompletion: { _ in
+                    Installer.install(ipaUrl: url, export: false, returnCompletion: { _ in
                     Task { @MainActor in
                         AppsVM.shared.fetchApps()
                         NotifyService.shared.notify(
--- a/PlayCover/Utils/URLHandler.swift
+++ b/PlayCover/Utils/URLHandler.swift
@@ -34,8 +34,8 @@ struct URLHandler {
 
     func processURL(url: URL) {
         guard let urlComponenents = NSURLComponents(url: url, resolvingAgainstBaseURL: false),
-            let uriPath = urlComponenents.path,
-            let params = urlComponenents.queryItems else {
+              let uriHost = urlComponenents.host,
+              let params = urlComponenents.queryItems else {
                 // Fall back to old url handler (for files)
                 if url.pathExtension == "ipa" {
                     Installer.install(ipaUrl: url, export: false, returnCompletion: { _ in
@@ -52,9 +52,10 @@ struct URLHandler {
         // URI format: playcoverapp://<object>?action=<action>&<param>=<value>
         // Example: playcoverapp://source?action=add&url=https://homebrew.playcover.io
 
+        debugPrint(urlComponenents)
         // Switch case for main uri path
-        switch uriPath {
-        case "/source":
+        switch uriHost {
+        case "source":
             processSourceURL(params: params)
         default:
             // Print URL to console and break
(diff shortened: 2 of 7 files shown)


Depal1 (dismissed):
A viable obfuscation method needs to be figured out for the very particular case of an app targeting the url schema as an unwanted “PlayCover detection”. For more information there is a proposal for an impersonation scheme stated in the comments of https://github.com/PlayCover/PlayCover/pull/534. Having none might not pose an immediate threat to PlayCover compatibility, but we need to plan ahead since this could impact the app’s core functionality as jailbreak detection does.

P.D. I will be able to resume work here in about a week.

amirsaam:
can we have a query in url schema for sources to add the source directly to avoid json link been copied and opened in browser?

ohaiibuzzle:
> A viable obfuscation method needs to be figured out for the very particular case of an app targeting the url schema as an unwanted “PlayCover detection”. For more information there is a proposal for an impersonation scheme stated in the comments of #534. Having none might not pose an immediate threat to PlayCover compatibility, but we need to plan ahead since this could impact the app’s core functionality as jailbreak detection does.
> 
> P.D. I will be able to resume work here in about a week.

This would (probably) require something fairly silly (like, modifying Info.plist while PlayCover is running) that I am not sure will work. Difficult at best

Cartfile.resolved
```diff
@@ -1 +1 @@
-github "PlayCover/PlayTools" "3.0.0-staging"
```
TheMoonThatRises: Cartfile change accidentally pushed here.

PlayCover/Utils/URLHandler.swift
```diff
@@ -0,0 +1,93 @@
...
+                    }})
+                }
+                return
+            }
+        // URI format: playcoverapp://<object>?action=<action>&<param>=<value>
+        // Example: playcoverapp://source?action=add&url=https://homebrew.playcover.io
+
+        // Switch case for main uri path
+        switch uriPath {
+        case "/source":
```
Depal1: Judging by the example above, I don't think the slash is intended in this case. Otherwise the url handler for the type source would have to be called with three slashes `playcover:///source?action=add&url=<source_url>`
```suggestion
        case "source":
```

PlayCover/Utils/URLHandler.swift
```diff
@@ -0,0 +1,93 @@
...
+            // Print URL to console and break
+            print("Unknown URL: \(url)")
+        }
+    }
+
+    func processSourceURL(params: [URLQueryItem]) {
+        switch params[0].value {
+        case "add":
+            // Add source
+            if let url = params[1].value {
```
Depal1: There is missing error handling if the request for actions includes no parameters. For instance, calling `playcover:///source?action=add` will crash the app due to an index out of range error.

Depal1: A similar situation occurs on the switch of line 66 if `playcover:///source?` is queried.

Depal1 (changes requested):

PlayCover/Utils/URLHandler.swift
ohaiibuzzle: Actually, no. Keep in mind this is the format of a URI
`URI = scheme ":" ["//" authority] path ["?" query] ["#" fragment]`
If you only have `source`, the URL will actually be
`playcover:source?action=add&url=something`
`/source`, meanwhile, will actually make
`playcover:/source?action=add&url=something`
Anyway, will update it

PlayCover/Utils/URLHandler.swift
ohaiibuzzle: Added the necessary checks

Depal1: The crashes still happen on https://github.com/PlayCover/PlayCover/pull/538/commits/9db78398d4079f087deb9148bc25ae66f9d48cc5. I propose merging the following changes to solve the issue https://github.com/Depal1/PlayCover/compare/9db7839...804016c.

PlayCover/Utils/URLHandler.swift
```diff
@@ -0,0 +1,92 @@
...
+                    URLObservable.shared.url = url
+                    URLObservable.shared.action = .add
+                }
+            case "remove":
+                // Remove source
+                if let url = params[1].value {
+                    URLObservable.shared.url = url
+                    URLObservable.shared.action = .remove
+                }
+            case "update":
```
amirsaam, line 89: In the matter of `remove` and `update`, I don't see any use case that ends well. because anybody can pass a remove url and delete my source for example in users' playcover (or vice versa) or even i can pass decrypt.day url as my old url and tell playcover to replace it with mine.
i know that user must accept any changes but at actually is more harmful than useful.
i would say we don't need such vulnerabilities since all users are not expertise and they usually don't check what's going on when they trust their providers.
`add` is enough. that's all.

ohaiibuzzle: That's reserved for future use. If you check, there is no code that makes use of `update` and `remove`. That is just prebuilt infrastructure in case we want to use it for other stuff in the code later.

amirsaam: since we can never guarantee that it won't be abused, i would say to not put it there in the codes at all

ohaiibuzzle: Again, it is added there as ready-built infrastructure. Other content (eg. keymaps or apps) can use it later on if needed (which they very well may be). It being there doesn't mean it will get implemented

PlayCover/Utils/URLHandler.swift
ohaiibuzzle: Should be fine now. Added one more check to make sure the params is what we expect

Depal1 (approved):

PR #554, opened 2022-12-06
Switch from optool to inject

IsaacMarovitz:


Depal1 (approved):
For code injection beyond PlayTools, it would be necessary to expose the inject cli for use on Terminal. This can be done in a later PR since the current one is self contained, and it replaces the functionalities of optool correctly.

PR #571, opened 2022-12-13
Feat: Define More Data in Sources

amirsaam:
This PR adds more data required from Sources

- [x] noPlayTools `Boolean` as a data to tell PlayCover Installer whether PlayTools should be installed into the app or not
- [ ] disabledSIP `Boolean` that informs user whom tries to install an app that works only in SIP disabled environment

Better to be not merged until `2.1.0` release as it will change Source Structure JSON


PlayCover/AppInstaller/Installer.swift
```diff
@@ -36,13 +36,15 @@ class Installer {
...
     // swiftlint:disable function_body_length
-    static func install(ipaUrl: URL, export: Bool, returnCompletion: @escaping (URL?) -> Void) {
+    static func install(ipaUrl: URL, export: Bool, ipaSourcePTD: Bool?, returnCompletion: @escaping (URL?) -> Void) {
         // If (the option key is held or the install playtools popup settings is true) and its not an export,
         //    then show the installer dialog
         let installPlayTools: Bool
 
         if (Installer.isOptionKeyHeld || InstallPreferences.shared.showInstallPopup) && !export {
             installPlayTools = installPlayToolsPopup()
+        } else if !(ipaSourcePTD ?? true) {
```
TheMoonThatRises, line 47: It would be better to ensure `ipaSourcePTD` to not be null before evaluating it instead of giving it a default value of `true`. Currently, this would make the last else statement unable to run if the app is installed via default means, and making the app always install without PlayTools despite the user's settings.

amirsaam: I have tested that, it does not do that. last else statement works fine because in source if you define `true` or in other installers like app library `nil` it will be skipped to `else` part, only works while it's `false` from source.

ohaiibuzzle:
First, probably you should exclude the Xcode project files from your commits 
Second, you probably should have make PlayTools field optional, because else this would probably break existing sources that don't provide that key in their JSON.

amirsaam:
> First, probably you should exclude the Xcode project files from your commits

Those are revert commits, I accidentally pushed them so I had to revert them manually

> Second, you probably should have make PlayTools field optional, because else this would probably break existing sources that don't provide that key in their JSON.

You mean in `StoreAppData` struct?

ohaiibuzzle:
> You mean in `StoreAppData` struct?

Yeah. Make it optional, or initialize it by default with `false`, just in case the user is adding a source that has not been updated yet. Those shouldn't cause exceptions, else this would be considered an API-incompatible change and deferred until next major release (per semver)

PR #599, opened 2022-12-20
Feat: add invisible app alias

Depal1:
- Create and remove "hidden" aliases for your apps under `App Settings > Misc`. This will in turn make PlayCover apps easily available from the Launchpad, as well as launching them with a game controller. If an app is to be deleted, its alias will be removed too.


ZhichGaming:
I tested this slightly, and it worked pretty well. Creating and removing alias worked and toggling keymapping in settings applied without even needing to launch the app through PlayCover. I have found one bug, and it is that the create/remove alias text does not change immediately after clicking it. You need to unfocus the app and refocus it for it to update to the correct state.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -460,6 +463,13 @@ struct MiscView: View {
                         }
                     }
                     Spacer()
+                    Button((app.hasAlias() == true) ? "settings.removeAlias" : "settings.createAlias") {
+                        if app.hasAlias() == false {
+                            shell.createAlias(app)
```
amirsaam, line 470: I think updating `hasAlias: Bool?` here to true and to false in else statement will fix the issue @ZhichGaming mentioned.

TheMoonThatRises: And also using it in the button instead of `(app.hasAlias() == true)`. Also it doesn't need an `== true` since it returns a bool.

Depal1:
>  I have found one bug, and it is that the create/remove alias text does not change immediately after clicking it. You need to unfocus the app and refocus it for it to update to the correct state.

This issue was addressed earlier today.

Depal1:
There is a remaining `DispatchQueque` which I will substitute later.

PlayCover/Model/PlayApp.swift
```diff
@@ -138,6 +155,12 @@ class PlayApp: BaseApp {
         Uninstaller.clearExternalCache(info.bundleIdentifier)
     }
 
+    func removeAlias() {
+        if FileManager.default.fileExists(atPath: aliasURL.path) {
```
TheMoonThatRises: `FileManager.default.delete` already checks if the file exists, so this would be a redundant check.

PlayCover/Model/PlayApp.swift
```diff
@@ -107,6 +107,19 @@ class PlayApp: BaseApp {
         }
     }
 
+    var aliasDirectory: URL {
```
TheMoonThatRises: Instead of making this a computed property, it can be a static constant, which could be easier to read and use.

PlayCover/Model/PlayApp.swift
```diff
@@ -107,6 +107,19 @@ class PlayApp: BaseApp {
...
+        return FileManager.default.homeDirectoryForCurrentUser
+            .appendingPathComponent("Applications")
+            .appendingPathComponent("PlayCover")
+    }
+
+    var aliasURL: URL {
+        return FileManager.default.homeDirectoryForCurrentUser
```
TheMoonThatRises: Since the first part of the url is the same, it might as well use `aliasDirectory` and be a lazy variable.

PlayCover/Utils/Shell.swift
```diff
@@ -138,6 +138,29 @@ class Shell: ObservableObject {
         }
     }
 
+    static func createAlias(_ app: PlayApp) {
+        if !FileManager.default.fileExists(atPath: app.aliasDirectory.path) {
```
TheMoonThatRises: This check is not necessary, it should be safe to remove this.

PlayCover/Utils/Shell.swift
```diff
@@ -138,6 +138,28 @@ class Shell: ObservableObject {
         }
     }
 
+    static func createAlias(_ app: PlayApp) {
+        do {
+            try FileManager.default.createDirectory(atPath: PlayApp.aliasDirectory.path,
+                                                    withIntermediateDirectories: true,
+                                                    attributes: nil)
+        } catch {
+                Log.shared.log("Error creating app alias directory: \(error)")
```
TheMoonThatRises: Indentation.

TheMoonThatRises (approved):
There is an extreme edge case where if you spammed the button, multiple aliases would be created, but I'm not sure how that could be prevented. Other than that LGTM.

Depal1:
> There is an extreme edge case where if you spammed the button, multiple aliases would be created, but I'm not sure how that could be prevented. Other than that LGTM.

I can think of `closeView.toggle()` as it is done with the PlayTools injection and removal button. Would it be convenient to use it in this case?

TheMoonThatRises:
> I can think of `closeView.toggle()` as it is done with the PlayTools injection and removal button. Would it be convenient to use it in this case?

That could work.

IsaacMarovitz:
This can be achieved programmatically with this:
https://developer.apple.com/documentation/foundation/nsurl/1408532-writebookmarkdata
https://stackoverflow.com/questions/1619073/how-do-you-create-an-alias-in-cocoa-is-a-symlink-good-enough

Theres a similar API for URL:
https://developer.apple.com/documentation/foundation/url/1779742-writebookmarkdata

Depal1:
Closed in favour of https://github.com/PlayCover/PlayCover/pull/738.

PR #622, opened 2022-12-24
Better Download Handler (Class)

amirsaam:
A new DownloadApp class has been introduced into PlayCover.
Able to cancel downloads, retries download as how many time you config it to do (default is 3)

Note: The DownloadManager package can handle multiple downloads at the same time and cancel each one of them if user wants but I avoided to implement that into PlayCover for multiple reasons like Installer.

DownloadApp class usage example:
```swift
if downloadVM.downloading {
    DownloadApp(url: nil, app: nil, warning: nil).cancel()
} else {
    DownloadApp(url: DownloadURL, app: StoreAppDataObject, warning: VersionCompareWarning).start()
}
```


PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -28,15 +29,24 @@ struct StoreAppView: View {
                                 isList: isList,
                                 warningSymbol: $warningSymbol,
                                 warningMessage: $warningMessage)
-        .gesture(TapGesture(count: 2).onEnded {
```
IsaacMarovitz, line 33: This gesture should be added back. Currently, the only way to download is through the context menu

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -28,15 +29,24 @@ struct StoreAppView: View {
...
-        })
         .environmentObject(downloadVM)
+        .contextMenu {
+             Button {
+                 if let url = URL(string: app.link) {
+                     let downloader = DownloadApp(url: url,
+                                                  app: app,
+                                                  warning: warningMessage)
+                     if downloadVM.downloading {
+                         downloader.cancel()
```
IsaacMarovitz: Cancel does not work the first time I try to select it.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -28,15 +29,24 @@ struct StoreAppView: View {
...
-        .gesture(TapGesture(count: 2).onEnded {
-            if let url = URL(string: app.link) {
-                downloadApp(url, app)
-            }
-        })
-        .simultaneousGesture(TapGesture().onEnded {
-            selected = app
-        })
         .environmentObject(downloadVM)
+        .contextMenu {
```
IsaacMarovitz: I'm not sure a context menu is the best UI choice here. The cancel button should be more visible i.e. on the download toast itself.

PlayCover/en.lproj/Localizable.strings
```diff
@@ -25,6 +25,7 @@
 "ipaLibrary.version.newer" = "This version is newer than the version you have installed";
 "ipaLibrary.alert.download" = "Are you sure you want to download this version of %@?";
 "ipaLibrary.download" = "Download";
+"ipaLibrary.cancelDL" = "Cancel Current Download";
```
IsaacMarovitz: Change this to `ipaLibrary.cancelDownload`

PlayCover/AppInstaller/Downloader.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.app = app
+        self.warning = warning
+    }
+
+    let downloadVM = DownloadVM.shared
+    let installVM = InstallVM.shared
+    let downloader = DownloadManager.shared
+
+    func start() {
+        if downloadVM.downloading && installVM.installing {
```
IsaacMarovitz: IMO you should add a `NetworkVM` check here

PlayCover/AppInstaller/Downloader.swift
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    private func proceedInstall(_ url: URL?) {
+        if let url = url {
+            uif.ipaUrl = url
+            Installer.install(ipaUrl: uif.ipaUrl!, export: false, returnCompletion: { _ in
+                DispatchQueue.main.async {
+                    FileManager.default.delete(at: url)
+                    AppsVM.shared.apps = []
```
IsaacMarovitz: Unnecessary, `fetchApps()` already clears the array

IsaacMarovitz (changes requested):

PlayCover/AppInstaller/Downloader.swift
amirsaam: It was in your own function, I suspect it would be necessary because if user press download and open AppLibrary after and wait for download there the list won't update. If it's not like that, tell me which line you want to be removed.

PlayCover/AppInstaller/Downloader.swift
amirsaam: Hmm, ok.

PlayCover/Views/App Views/StoreAppView.swift
amirsaam: Known issue, I have no fix right now.
I would love to have some help on it.

PlayCover/Views/App Views/StoreAppView.swift
amirsaam: Is it ok to it just start download?

PlayCover/Views/App Views/StoreAppView.swift
amirsaam: fixed the issue btw

IsaacMarovitz (approved):

Depal1 (approved):

Keys222:
Download play cover

PR #588, opened 2022-12-17
Fancy IPA Library + New Downloader + New Cache System

amirsaam:
Co-Authored-By: Isaac Marovitz <42140194+IsaacMarovitz@users.noreply.github.com>

This PR introduces several changes:

- New detailed page for each app in IPA Libraey based on iTunesLookup file that its function is already being used and handles fetching app icons in IPA Library (No extra function is required) by @IsaacMarovitz
- New `StackNavigationView` view extension has been introduced in `ViewExtenstions.swift` that is similar to `NavigationStack` on macOS 13 but supports macOS 12, so DetailedView is accessible in macOS 12 with same code. This extension can be used in any other views as you may want like creating App Folders in App Library or Source Folders in IPA Library
- New `DownloadApp` class has been introduced in `Downloader.swift` explained [here](https://github.com/PlayCover/PlayCover/pull/588#issuecomment-1361166253)
- Added working Download or Cancel Download buttons in the detailed page of each app
- Download buttons will show 4 different text:
1. `Get`: If the user hasn't installed that app/game yet on their PlayCover
2. `Reinstall`: If the user already has the same version of the app installed in their PlayCover
3. `Update`: If user has older version compared to what they are viewing installed in their PlayCover
4. `Downgrade`: If user has newer version compared to what they are viewing installed in their PlayCover

Note: This version compare is using same logic and function of what @TheMoonThatRises has introduced into PlayCover in previously merged PRs.

- Download buttons on press will change to Download progress bar with a stop option and after download to Install progress bar and will be shown only when user is in that app detail page
- Made showing ToastView optional that can be controlled via `ToastVM.shared.isShown` boolean, it will no longer be shown when downloading or installing an app while user is in that app's DetailedView

Note: If you navigate to other apps DetailedView you'll still see the 4 type of download texts but they are disabled, no matter what in any DetailedView while app is being installed, that button is disabled.

- Apps that have no iTunesLookup will get default text that notifies user the app is not on app store
- New `Cacher` class has been introduced in `Cacher.swift` that uses `DataCache` package, removed `ImageCache` class and rewrote the logic.
- Online Images now are being cached via `CachedAsyncImage` package that uses `URLCache` as memory 
- All texts have locale key, so translation can start right away

This library in future PlayCover development can be used to shown extra details that can be provided by sources like a stat badge like file size that shows if the app should have PlayTools or not or it just works in SIP disabled environment or not

[...]

PR #625, opened 2022-12-24
Fancy IPA Library

amirsaam:
_Reworked Fancy IPA Library_
**THIS PR NO LONGER REQUIRES MACOS 13 AS PREVIOUSLY NEEDED**

All descriptions already have been given in #588 
_Separated that PR into multiple parts_

- `AppDetailedView` base design is from @IsaacMarovitz
- I reworked it, fixed issues, improved its compatibility, added download states, added locale and etc.

_THIS NOTE SHOULD BE READ:_

- New `StackNavigationView` view extension has been introduced in `ViewExtenstions.swift` that is similar to `NavigationStack` on macOS 13 but supports macOS 12.
- So `DetailedView` is accessible in macOS 12 with the same code with no `if #available` statement.

This extension can be used in any other views as you may want.
For example: Creating App Folders in App Library or Source Folders in IPA Library

Co-Authored-By: Isaac Marovitz <42140194+IsaacMarovitz@users.noreply.github.com>


IsaacMarovitz:
Needs rebase

amirsaam:
> Needs rebase

Elaborate please.

PR #653, opened 2022-12-29
Add `force_unwrapping` rule to SwiftLint

IsaacMarovitz:
**TODO:**
- [ ] Split view detection function is currently broken

### DIFF
--- a/PlayCover/Views/MainView.swift
+++ b/PlayCover/Views/MainView.swift
@@ -201,8 +201,10 @@ struct MainView: View {
                             .tint(Color.gray)
                             .controlSize(.large)
                             Button("xcode.install.failed.altButton") {
-                                NSWorkspace.shared.open(URL(string:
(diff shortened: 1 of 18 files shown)


PlayCover/Views/MainView.swift
```diff
@@ -285,8 +287,12 @@ struct SplitViewAccessor: NSViewRepresentable {
             var sview = superview
 
             // Find split view through hierarchy
-            while sview != nil, !sview!.isKind(of: NSSplitView.self) {
-                sview = sview?.superview
+            if var sview = sview {
```
TheMoonThatRises, line 290: Split view detection can be fixed by replacing this bit of code with
```swift
while let nsview = sview, !nsview.isKind(of: NSSplitView.self) {
    sview = sview?.superview
}
```
since `sview` needs to be checked for `nil` every loop instead of just the first.

PR #654, opened 2022-12-30
Informative Install Infos

amirsaam:
Needs #625 

Adds an Optional Array to Source that contains three Bools

Test Json: https://amrsm.ir/test.json

1. `disabledSIP`: If `true` means app/game only works while sip is disabled
2. `noPlayTools`: if `true` means app/game should not or better not have playtools installed
3. `signingSetup`: if `true` means app/game needs to do amfi thing to login to account
Screenshot 1401-10-25 at 22 21 45
Screenshot 1401-10-25 at 22 21 49


ohaiibuzzle:
I'd say instead of being a section in the App Info panel, it should be a dialog after clicking on Install but before you proceed.
This way it will be in the face of people and those who aren't reading, we'll know

amirsaam:
@ohaiibuzzle In that way user always needs an extra unnecessary step to access these datas. Currently user always has access to these when he comes into detail page even after install or just if sip disabled was required they can decide to click on download or not.

ohaiibuzzle:
> @ohaiibuzzle In that way user always needs an extra unnecessary step to access these datas. Currently user always has access to these when he comes into detail page even after install or just if sip disabled was required they can decide to click on download or not.

It's an extra step of course, but it would avoid the trouble of people just not reading the instructions and go to conclusion, or ask questions that are already answered. Kind of like how Google Play notifies you of app permissions before installing. It forces people to actually read

amirsaam:
> > @ohaiibuzzle In that way user always needs an extra unnecessary step to access these datas. Currently user always has access to these when he comes into detail page even after install or just if sip disabled was required they can decide to click on download or not.
> 
> It's an extra step of course, but it would avoid the trouble of people just not reading the instructions and go to conclusion, or ask questions that are already answered. Kind of like how Google Play notifies you of app permissions before installing. It forces people to actually read

I'll consider adding an alert to be triggered while user is not in detailed view and wants to download the app (like right click download in ipa library) but not in detailed view itself, that's too annoying tbh

PR #656, opened 2022-12-30
Primary App Icon Default

TheMoonThatRises:
Always find the primary image icon as a backup in case the icon name is not `AppIcon`.


PlayCover/Utils/Cacher.swift
```diff
@@ -27,6 +27,18 @@ class Cacher {
...
+            try app.url.enumerateContents { file, _ in
+                if file.lastPathComponent.contains(app.info.primaryIconName), let icon = NSImage(contentsOf: file),
+                    checkImageDimensions(icon, bestResImage) {
+                    bestResImage = icon
+                }
+            }
+        } catch {
+            Log.shared.error(error)
+        }
+
```
amirsaam, line 41: I'm confused with giving `bestResImage` a value twice.

IsaacMarovitz (approved):

PR #658, opened 2022-12-30
Removal of old cache clear command

amirsaam:
Since I tested in nightly and clearing old ImageCache folder sends error, there is no point for having it.

### DIFF
--- a/PlayCover/Views/MenuBarView.swift
+++ b/PlayCover/Views/MenuBarView.swift
@@ -96,14 +96,6 @@ struct PlayCoverViewMenuView: Commands {
             Button("menubar.clearCache") {
                 DataCache.instance.cleanAll()
                 URLCache.iconCache.removeAllCachedResponses()
-                do {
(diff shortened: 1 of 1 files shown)


PlayCover/Views/MenuBarView.swift
```diff
@@ -96,14 +96,6 @@ struct PlayCoverViewMenuView: Commands {
             Button("menubar.clearCache") {
                 DataCache.instance.cleanAll()
                 URLCache.iconCache.removeAllCachedResponses()
-                do {
-                    let oldCacheFolder = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
```
TheMoonThatRises, line 100: The reason why the old clear cache command did not work as intended is because it lacked two lines:
```swift
         .appendingPathComponent(Bundle.main.bundleIdentifier!)
         .appendingPathComponent("Image Cache")
```
Instead, this function is trying to remove the entire cache directory (`~/Library/Caches`). So, those two lines should be added at the end to make the function act as intended.

PR #664, opened 2023-01-02
Fix: Genshin account switcher now works.

JoseMoreville:
This was a bug introduced by the refactor of the code, fixed by removing folderName from the path for 'Storage' folder.


Depal1 (approved):

AJ5735:
How to fix it?

PR #710, opened 2023-01-17
PlayChain keychain emulation support

ohaiibuzzle:
Needs https://github.com/PlayCover/PlayTools/pull/65

Add PlayChain toggles in the PlayCover UI

### DIFF
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -21,9 +21,32 @@ struct AppSettingsData: Codable {
     var notch: Bool = NSScreen.hasNotch()
     var bypass = false
     var discordActivity = DiscordActivity()
-    var version = "2.0.0"
+    var version = "3.0.0"
     var playChain = false
     var playChainDebugging = false
+
+    init() {}
+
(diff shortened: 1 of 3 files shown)


PlayCover/Model/AppSettings.swift
```diff
@@ -22,6 +22,8 @@ struct AppSettingsData: Codable {
     var bypass = false
     var discordActivity = DiscordActivity()
     var version = "2.0.0"
+    var playChain = false
```
TheMoonThatRises, line 25: Since two new keys are being added to the settings file, there should be a decoder function similar to how keymapping does it to provide backward compatibility and prevent erasure of settings set before these keys were added.

ohaiibuzzle: Done!

Depal1 (approved):

Depal1:
PlayChain should be added to the uninstaller in a future PR.

00UNPRO00:
How to fix this? And also, where can I find the settings to enable Playchain? Thank you

Screenshot 2023-04-03 at 21 07 46

PR #711, opened 2023-01-17
Remove Xcode CLI Dependancy

IsaacMarovitz:
**Change:**
- Replaces all uses of Xcode CLI with our own implementations, based on `inject`
- Installing Xcode CLI would no longer be required after this PR

`install_name_tool` usage not converted cause it's not necessary afaik

I will leave removing the UI and backend surrounding Xcode CLI to another PR so that the changes from this one can easily be tested in isolation.

### DIFF
--- a/PlayCover/Utils/DataExtensions.swift
+++ b/PlayCover/Utils/DataExtensions.swift
@@ -1,0 +1,29 @@
+//
+//  DataExtensions.swift
+//  PlayCover
+//
+
+import Foundation
+
+extension String {
+    init(data: Data, offset: Int, commandSize: Int, loadCommandString: lc_str) {
+        let loadCommandStringOffset = Int(loadCommandString.offset)
+        let stringOffset = offset + loadCommandStringOffset
+        let length = commandSize - loadCommandStringOffset
+        self = String(data: data[stringOffset..<(stringOffset + length)],
+                      encoding: .utf8)!.trimmingCharacters(in: .controlCharacters)
+    }
+}
+
+extension Data {
+    func extract<T>(_ type: T.Type, offset: Int = 0) -> T {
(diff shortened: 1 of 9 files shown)


IsaacMarovitz:
Needs further testing but heres a list of apps I've tried (all have worked):
- Genshin Impact (big, modern)
- Terraria (modern, fat)
- Payback2 (fat)
- Temple Run (old)
- Plants v.s. Zombies (old)
- Snapchat (modern)

IsaacMarovitz:
> i've tested all the apps in the example source
- Moon

IsaacMarovitz:
download

PlayCover/Utils/PlayTools.swift
```diff
@@ -255,32 +538,33 @@ class PlayTools {
...
-        get throws {
-            try binPath("vtool")
-        }
-    }
-
-    private static var otool: URL {
-        get throws {
-            try binPath("otool")
-        }
+extension URL {
```
TheMoonThatRises: It would be best to move URL and String extensions into the URLExtensions file along with the other URL and String extensions.

TheMoonThatRises (approved):

PlayCover/Utils/DataExtensions.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  DataExtensions.swift
+//  PlayCover
+//
+
+import Foundation
+
+extension String {
```
TheMoonThatRises, line 8: Theres a String extension in the URLExtensions file, which you might want to move here, or move this definition there.

IsaacMarovitz: The String extension in the URLExtensions file is relevant to URLs

Depal1 (approved):

PR #731, opened 2023-01-21
Fix wrong German translation

SkyrilHD:
There is a wrong translation for the word 'Finder'.

Sucher means 'seeker' As Finder is a name, there is no need for translation.


zanderp25:
We use Weblate to manage translations, so you should fix it there. 

https://hosted.weblate.org/projects/playcover/playcover/

PR #732, opened 2023-01-22
Animation Fix and Revert back to my Custom NavigationStack

amirsaam:


PlayCover/Utils/NavigationStack.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    case none, defaultTranisition, custom(AnyTransition)
+
+    var id: String {
+        switch self {
+        case .none: return "none"
+        case .defaultTranisition: return "default"
+        case .custom: return "custom"
+        }
+    }
```
KhoraLee, line 23:
```suggestion
    static func == (lhs: StackNavigationTransition, rhs: StackNavigationTransition) -> Bool {
        switch (lhs, rhs) {
        case (.none, .none):
            return true
        case (.defaultTranisition, .defaultTranisition):
            return true
        case (.custom, .custom):
            return true
        default:
            return false
        }
    }

    case none, defaultTranisition, custom(AnyTransition)
```

PlayCover/Utils/NavigationStack.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    init(currentSubview: Binding<AnyView>,
+         showingSubview: Binding<Bool>,
+         transition: StackNavigationTransition,
+         @ViewBuilder rootView: @escaping () -> RootContent) {
+        self._currentSubview = currentSubview
+        self._showingSubview = showingSubview
+        self.transition = transition
+        self.rootView = rootView
+    }
```
KhoraLee, line 54: IIRC, `@escaping` is not good to use in SwiftUI. And making transition `@State` variable doesn't seem necessary.
```suggestion
    let rootView: RootContent
    let transition: StackNavigationTransition

    @Binding var currentSubview: AnyView
    @Binding var showingSubview: Bool

    init(currentSubview: Binding<AnyView>,
         showingSubview: Binding<Bool>,
         transition: StackNavigationTransition,
         @ViewBuilder rootView: () -> RootContent
    ) {
        self.rootView = rootView()
        self.transition = transition
        self._currentSubview = currentSubview
        self._showingSubview = showingSubview
    }
```

PlayCover/Utils/NavigationStack.swift
```diff
@@ -0,0 +1,114 @@
...
+                                ? .move(edge: .leading)
+                                : transition.anyTransition)
+            } else {
+                StackNavigationSubview(currentSubview: $currentSubview,
+                                       isVisible: $showingSubview,
+                                       transition: transition) {
+                    currentSubview
+                        .transition(transition.anyTransition)
+                        .zIndex(1)
+                }
```
KhoraLee, line 71: Is there any reason declaring and using StackNavigationSubview?

PlayCover/Utils/NavigationStack.swift
amirsaam: Done!

PlayCover/Utils/NavigationStack.swift
amirsaam: Thanks!

PlayCover/Utils/NavigationStack.swift
amirsaam: Actually I don't remember why I did it but now I see it's redundant as you said. Thanks for mentioning it!

PR #740, opened 2023-01-22
Feat: add invisible app alias

Depal1:
A reimplementation of https://github.com/PlayCover/PlayCover/pull/599 with `writeBookmarkData` instead of AppleScript.

### DIFF
--- a/PlayCover/Model/PlayApp.swift
+++ b/PlayCover/Model/PlayApp.swift
@@ -157,7 +157,7 @@ class PlayApp: BaseApp {
                                                 includingResourceValuesForKeys: nil, relativeTo: nil)
             try URL.writeBookmarkData(data, to: aliasURL)
         } catch {
-            Log.shared.log("\(error)")
+            Log.shared.log(error.localizedDescription)
         }
     }
 
--- a/PlayCover/Model/PlayApp.swift
+++ b/PlayCover/Model/PlayApp.swift
@@ -107,6 +107,12 @@ class PlayApp: BaseApp {
         }
     }
 
+    static let aliasDirectory = FileManager.default.homeDirectoryForCurrentUser
+            .appendingPathComponent("Applications")
+            .appendingPathComponent("PlayCover")
+
+    lazy var aliasURL = PlayApp.aliasDirectory.appendingPathComponent(name)
+
     lazy var settings = AppSettings(info, container: container)
 
     lazy var keymapping = Keymapping(info, container: container)
@@ -120,6 +126,10 @@ class PlayApp: BaseApp {
             Log.shared.error(error)
             return true
         }
+    }
+
+    func hasAlias() -> Bool {
+        return FileManager.default.fileExists(atPath: aliasURL.path)
     }
 
     func isCodesigned() throws -> Bool {
@@ -136,6 +146,28 @@ class PlayApp: BaseApp {
 
     func clearAllCache() {
         Uninstaller.clearExternalCache(info.bundleIdentifier)
+    }
+
+    func createAlias() {
+        do {
+            try FileManager.default.createDirectory(atPath: PlayApp.aliasDirectory.path,
+                                                    withIntermediateDirectories: true,
+                                                    attributes: nil)
+        } catch {
+            Log.shared.log("Error creating app alias directory: \(error)")
+        }
+
+        do {
+            let data = try url.bookmarkData(options: .suitableForBookmarkFile,
+                                                includingResourceValuesForKeys: nil, relativeTo: nil)
+            try URL.writeBookmarkData(data, to: aliasURL)
+        } catch {
+            Log.shared.log("Error creating bookmark data or writing to destination file: \(error)")
+        }
+    }
+
+    func removeAlias() {
+        FileManager.default.delete(at: aliasURL)
     }
 
     func deleteApp() {
--- a/PlayCover/Model/PlayApp.swift
+++ b/PlayCover/Model/PlayApp.swift
@@ -154,7 +154,7 @@ class PlayApp: BaseApp {
                                                     withIntermediateDirectories: true,
                                                     attributes: nil)
         } catch {
-            Log.shared.log("Error creating app alias directory: \(error)")
+            Log.shared.log("\(error)")
         }
 
         do {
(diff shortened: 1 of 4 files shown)


PlayCover/Views/Uninstaller.swift
```diff
@@ -123,6 +123,7 @@ class Uninstaller {
             FileManager.default.delete(at: app.entitlements)
         }
 
+        app.removeAlias()
```
TheMoonThatRises, line 129: The directory `~/Applications/PlayCover` should be added to the `pruneURLs` array and the `removeAlias` should be a new uninstall setting for cases like uninstalling and reinstalling apps.

Depal1: I added the directory to another array since it has to be indexed by display name instead of bundle id.

As for the second point, I have tested updating and reinstalling apps, and the existing aliases will still work for the newly replaced apps. I don't think we should leave an alias for inexistent references (deleted apps), so removing them should be default. The only way an alias can stay if an app is uninstalled, is if the app itself is deleted from Finder and not PlayCover. If that happens, using prune will take care of the orphan references.

PlayCover/Model/PlayApp.swift
```diff
@@ -138,6 +148,28 @@ class PlayApp: BaseApp {
         Uninstaller.clearExternalCache(info.bundleIdentifier)
     }
 
+    func createAlias() {
+        do {
+            try FileManager.default.createDirectory(atPath: PlayApp.aliasDirectory.path,
+                                                    withIntermediateDirectories: true,
+                                                    attributes: nil)
+        } catch {
+            Log.shared.log("Error creating app alias directory: \(error)")
```
TheMoonThatRises: These error should be localised or just have `Log.shared.log` output `error` itself. If the error is just going to be logged by itself, it should be combined with the do/catch statement below.

PlayCover/Views/Uninstaller.swift
```diff
@@ -149,5 +154,18 @@ class Uninstaller {
                 Log.shared.error(error)
             }
         }
+
+        for url in otherPruneURLs {
+            do {
```
TheMoonThatRises: Also I think the do/catch statement can be put outside the for loop and be combined with the do/catch statement above (the `pruneURLs` for loop).

TheMoonThatRises (changes requested):

PlayCover/Model/PlayApp.swift
Depal1: Done.

PlayCover/Views/Uninstaller.swift
Depal1: Done.

IsaacMarovitz:
Is it just me or is this not working at least not from the Xcode build

PlayCover/Model/PlayApp.swift
```diff
@@ -138,6 +148,28 @@ class PlayApp: BaseApp {
         Uninstaller.clearExternalCache(info.bundleIdentifier)
     }
 
+    func createAlias() {
+        do {
```
TheMoonThatRises, line 152: These do/catches should be converted to one, and there is no need for the quotation mark around the error.

Depal1:
> Is it just me or is this not working at least not from the Xcode build

I have built it again, and the aliases are created even if the directory does not exist. They will be in `~/Applications/PlayCover`, not `/Applications`.

PlayCover/Model/PlayApp.swift
Depal1: Removing the quotation marks or the back slash in the error will result in a type conversion error. Log.shared.log requires type String.

TheMoonThatRises (approved):

IsaacMarovitz:
> I have built it again, and the aliases are created even if the directory does not exist. They will be in `~/Applications/PlayCover`, not `/Applications`.

This isn't the right directory either:
Screenshot 2023-01-23 at 09 31 43
Where PlayCover is putting them

Screenshot 2023-01-23 at 09 31 53
Where my actual folder is

IsaacMarovitz:
You should use FileManager for this:
https://developer.apple.com/documentation/foundation/filemanager/searchpathdirectory/applicationdirectory

Depal1:
> > I have built it again, and the aliases are created even if the directory does not exist. They will be in `~/Applications/PlayCover`, not `/Applications`.
> 
> This isn't the right directory either: Screenshot 2023-01-23 at 09 31 43 Where PlayCover is putting them
> 
> Screenshot 2023-01-23 at 09 31 53
> 
> Where my actual folder is

That is intended since that is not your user folder. All other apps I know that use aliases put them in `~/Applications` (Steam, Crossover, RPCS3 for instance). The system directory is not meant to have aliases, specially since PlayCover apps reside in the user Library and not the system’s.

If the contrary where to happen, other users would see ‘apps’ (aliases) they won’t be able to access.

IsaacMarovitz:
> That is intended since that is not your user folder. All other apps I know that use aliases put them in `~/Applications` (Steam, Crossover, RPCS3 for instance). The system directory is not meant to have aliases, specially since PlayCover apps reside in the user Library and not the system’s.
> 
> If the contrary where to happen, other users would see ‘apps’ (aliases) they won’t be able to access.

I think that defeats the point of this feature, then. For the majority of users who only have one account on their system, this feature won't be particularly helpful like this.

IsaacMarovitz:
The reason I suggested using FileManger API is that it should do the best of both worlds, ie. use System Application folder where appropriate, and User Application folder when appropriate

Depal1:
> You should use FileManager for this:
> https://developer.apple.com/documentation/foundation/filemanager/searchpathdirectory/applicationdirectory

I missed this when I was writing my previous message. How does FileManager do this automatically? It would depend if either `.localDomainMask` or `.userDomainMask` is used.

Depal1:
> > That is intended since that is not your user folder. All other apps I know that use aliases put them in `~/Applications` (Steam, Crossover, RPCS3 for instance). The system directory is not meant to have aliases, specially since PlayCover apps reside in the user Library and not the system’s.
> > If the contrary where to happen, other users would see ‘apps’ (aliases) they won’t be able to access.
> 
> I think that defeats the point of this feature, then. For the majority of users who only have one account on their system, this feature won't be particularly helpful like this.

The purpose of this feature has always been Launchpad access.

IsaacMarovitz:
> The purpose of this feature has always been Launchpad access.

That should be specified in the button locale imo

IsaacMarovitz (approved):

PR #753, opened 2023-01-29
Add `force_unwrapping` rule to SwiftLint

IsaacMarovitz:
- Removes all force unwraps (except for a few in PlayTools and other places)
- Adds the `force_unwrapping` rule to SwiftLint
- Removes disabling of `inclusive_language` rule from SwiftLint
- Removes UIF
- Refactors Shell


PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -187,8 +187,8 @@ struct StoreAppConditionalView: View {
                 await Cacher().resolveITunesData(app.itunesLookup)
             }
             itunesResponce = try? cache.readCodable(forKey: app.itunesLookup)
-            if itunesResponce != nil {
-                onlineIcon = URL(string: itunesResponce!.results[0].artworkUrl512)
+            if let responce = itunesResponce {
```
JoseMoreville: There's a typo here

PlayCover/Utils/Shell.swift
```diff
@@ -23,30 +23,34 @@ class Shell: ObservableObject {
...
+                    }
+                    throw errOutput
                 }
-                throw errOutput
             }
-		}
-		return output
+            return output
+        } else {
+            return ""
```
JoseMoreville: I think we should send a different message here, instead of returning an empty string.

JoseMoreville: Its useful to add it but not @discardableResult may handle that

JoseMoreville (commented):
So far it looks good, great job isaac. just a few reviews

JoseMoreville (approved):

PR #766, opened 2023-01-31
Implement multithreaded downloader

KhoraLee:


TheMoonThatRises:
While you are changing the downloader, could you implement a download size warning? Like if the file size is, say above 1 GB, then it will warn the user before downloading?

amirsaam:
It would be lovely if instead of implement the downloader code directly you make it a package and implement that package into playcover. there is no decent swift multi thread downloader out there.
cheers and good job.

PlayCover/Views/Settings/DownloaderSettings.swift
```diff
@@ -0,0 +1,32 @@
...
+
+struct DownloaderSettings: View {
+    public static var shared = DownloaderPreferences()
+
+    @ObservedObject var downloaderPreferences = DownloaderPreferences.shared
+
+    var body: some View {
+        Form {
+            TextField("Chunk Size", value: downloaderPreferences.$chunkSize, formatter: NumberFormatter())
+            TextField("Max Connections", value: downloaderPreferences.$maxConnections, formatter: NumberFormatter())
```
amirsaam, line 26: These two better to be sliders I think in order to have min/max values.

KhoraLee:
I did test but looks like it has no difference in download speed with URLSession's downloadTask (URLSeesion is slightly faster..)
Tested with Genshin ipa(3.71GB)

Screenshot 2023-02-03 at 10 58 55 PM

PR #781, opened 2023-02-05
Adds toggle to fix screen issues with some apps in adaptive mode

JoseMoreville:
Basically a fix for some the bugs introduced to playcover after MacOS 13.2 depends on https://github.com/PlayCover/PlayTools/pull/77


PlayCover/Views/AppSettingsView.swift
```diff
@@ -272,6 +272,14 @@ struct GraphicsView: View {
                     }
                 }
                 HStack {
+                    let majorVersion = ProcessInfo().operatingSystemVersion.majorVersion
+                    let minorVersion = ProcessInfo().operatingSystemVersion.minorVersion
+                    let num = Float("\(majorVersion).\(minorVersion)")
+                    if num ?? 13.1 > 13.1 {
```
Depal1:
```suggestion
                    if #available(macOS 13.2, *) {
```

Depal1 (changes requested):
macOS version checks for features should be left to Swift's [check availability](https://developer.apple.com/documentation/swift/marking-api-availability-in-objective-c) conditionals.

Depal1 (approved):

PR #791, opened 2023-02-07
feat!: KeyCover keychain encryption for PlayChain

ohaiibuzzle:
This PR proposes a new set of feature for PlayCover called KeyCover that works in tandom with the previously introduced PlayChain
Since PlayChain stores keys as-is, an actor from macOS or an app installed in PlayCover can simply dump the chain of every PlayChain-enabled apps.
KeyCover mitigates this issue by encrypting the keys using AES-256-CBC with a key that is either set by the user or automatically generated and saved in Apple Keychain

Also includes some performance-related changes to PlayCover (eg. making app launch asynchronous)


PlayCover/Utils/KeyCover.swift
```diff
@@ -0,0 +1,313 @@
...
+    func wipeEncryptedKeyFile() throws {
+        try? FileManager.default.removeItem(at: encryptedKeyFile)
+    }
+}
+
+class KeyCoverMaster {
+    static let shared = KeyCoverMaster()
+
+    func setMasterKey(_ key: String) {
+        let tag = "com.playcover.masterkey"
```
TheMoonThatRises: Should this not be `io.playcover.masterkey` since PlayCover's domain is `.io`?

ohaiibuzzle: Oops. I forgot

IsaacMarovitz (approved):

PR #824, opened 2023-02-28
Allow user to suppress low power mode alert

### LINKED ISSUE #823: [Feature]: Settings for Power Saving Mode Notify.

My MacBook will turn to power saving mode when using battery. When I plug in and off several times, I need to click the notify of power saving mode several times.

Give a option to control the notify or let the notify only appear once.

### Anything else?

Nothing else.

### Issue Language

TheMoonThatRises:
Fixes #823.


Depal1 (approved):
It would be better to implement this with a 'don't remind me again'.

PR #851, opened 2023-03-09
xclt less installer macho converter code cleanup

XuYicong:
Created a new file `Macho.swift` to put macho-related code, so that `PlayTools` has less lines than `swiftlint`'s upper bound

Added sanity checks while iterating load commands.

Added byte order checks while replacing load commands.

Variable names changed to more meaningful ones.


IsaacMarovitz:
Code looks good, thank you for the cleanup! Gonna test the full suite of apps now just to make sure nothing went wrong.

IsaacMarovitz (approved):
Works perfectly

PR #861, opened 2023-03-12
Remove CoD:M Garena from Prohibition List

amirsaam:
Since we have reports that CoD:M Garena accounts won't get banned on PlayCover and there is a resolution issue that needs PlayTools to be ran for 1st time from PlayCover, there is no reason to let it still be prohibited.


JoseMoreville:
Allowing the game will be a boost for many players, it helps with aiming a whole lot and will be considered as cheating. This will be probably consider as cheating.

amirsaam:
Since the game will see the player uses Keyboard and Mouse they will be matched against same playerbase so there will be no real advantages.
We prohibited CoD cuase they banned players no matter what but Garena version won't do the same

Note: CoD has built-in Mouse and Keyboard support

IsaacMarovitz (approved):

PR #880, opened 2023-03-20
Feat: Source Folders & Priority + Optional Alphabetical Sort

amirsaam:
Introduces Source Folders in IPA Library where changes the structure of Sources and adds Name and Logo URL to them and shows them as they are.
With this approach we are no longer stack all apps from all sources together so for example we can have Genshin Impact from every source in their folders.
- [x] Optimised `StoreVM`
- [x] Needed UI Changes
- [x] Working Move Source Up/Down in the Setting
- [x] Backport Compatibility for older Sources
- [x] Add Search Apps Back
- [x] Fix UI Issues
- [x] Add Alphabetical Sort Option for Source Contents
- [x] Adds collapsible list of source folders under IPA Library in sidebar

**Note**:
There is a new package `Semaphore` used in this PR, this package introduces `AsyncSemaphore` alike `DispatchSemaphore` but unlike that, it does not block any thread; Instead, Swift concurrency tasks are suspended "awaiting" for the semaphore.

This is a good thing to have for future uses like I have done in this PR dealing with suspension in the easiest way.
The reason I did not use `try await Task.sleep` is that we aren't sure when the previous item in the loop is done to start the loop again, plus we are updating UI and we need to be on main thread and `DispatchSemaphore` will not work and sends PlayCover to Not Responding state.

**Issues**:
1. When users search apps, in meantime if they change sort option it will not change the app list until adds/removes a character from their search (just saying, I don't think it's a big deal)
2.  **FIXED** | Original issue:
```
The famous Search bar issue, if gets focused and back to root it will not get unfocused until clicked on the window but still will comeback after opening a source folder again until you switch once to App Library (I have no fix)
```


amirsaam:
New Source Struct + Old Struct Support:
Screenshot 1402-01-11 at 21 18 06
When opened a folder in IPA Library:
Screenshot 1402-01-11 at 21 21 33
When opened from sidebar:
Screenshot 1402-01-11 at 21 18 11

PlayCover/Utils/NavigationStack.swift
```diff
@@ -0,0 +1,111 @@
...
+//  NavigationStack.swift
+//  PlayCover
+//
+//  Created by Amir Mohammadi on 11/2/1401 AP.
+//
+
+import SwiftUI
+
+enum StackNavigationTransition: Equatable {
+    case none, defaultTranisition, custom(AnyTransition)
```
JoseMoreville: There is a typo here

PlayCover/Utils/NavigationStack.swift
```diff
@@ -0,0 +1,111 @@
+//
+//  NavigationStack.swift
+//  PlayCover
+//
+//  Created by Amir Mohammadi on 11/2/1401 AP.
+//
+
+import SwiftUI
+
+enum StackNavigationTransition: Equatable {
```
JoseMoreville, line 10: Is Equatable the best option to handle this operations?

PlayCover/ViewModel/NetworkVM.swift
```diff
@@ -7,6 +7,7 @@
 
 import SystemConfiguration
 import Foundation
+import Semaphore
```
JoseMoreville: Wasn't `import Semaphore` added before? I'm pretty much sure DispatchSemaphore can handle it without explicitly importing Semaphore

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -6,184 +6,192 @@
...
             }
         }
     }
+}
 
-    func appendSourceData(_ data: SourceData) {
-        self.sources.append(data)
-        self.resolveSources()
-    }
+// Source Data Structure
```
JoseMoreville, line 174: These structs would be better on a different file

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -6,184 +6,192 @@
...
-                if index > self.sources.endIndex {
-                    index = self.sources.endIndex
+    //
+    private func getSourceData(sourceLink: String) async -> (SourceJSON?, SourceValidation) {
+        guard let url = URL(string: sourceLink) else { return (nil, .badurl) }
+        do {
+            let (data, response) = try await URLSession.shared.data(
+                for: URLRequest(url: url, cachePolicy: .reloadIgnoringLocalAndRemoteCacheData)
+            )
+            if !url.isFileURL {
```
JoseMoreville, line 147: You should check for local files before starting any URLRequest

PlayCover/Views/Sidebar Views/IPALibraryView.swift
JoseMoreville, line 1: Can you please separate Sources view(Source Selection view) and Source library view (Apps of a source)

JoseMoreville (changes requested):
Left some comments that need to be checked

PlayCover/ViewModel/StoreVM.swift
amirsaam: Even if URL is local file we need to do URLRequest (or I am worng?)
So we are just checking afterwards to see if the url is local file and bypass the 200 statuscode check for it. nothing more

PlayCover/ViewModel/NetworkVM.swift
amirsaam: Actually that is a mistaken import there, DispatchSemaphore is still in authority to check the NetworkVM not AsyncSemaphore. I'll remove.

PlayCover/ViewModel/StoreVM.swift
JoseMoreville: As long it does not depend on response it's ok to move it, but there are better ways to check for local files, although the code works and the order is right, you should check for that before. just a suggestion.

PlayCover/Utils/NavigationStack.swift
amirsaam: Actually yes, and those checks were suggested by KhoraLee before here https://github.com/PlayCover/PlayCover/pull/732

PlayCover/ViewModel/StoreVM.swift
amirsaam: surely we need the `response` to be present to bypass it, right?

amirsaam: looks like it was a misunderstanding. resolving this.

PlayCover/ViewModel/StoreVM.swift
amirsaam: we have no file for data structs, it is somehow redundant to create one just for source structs; isn't it?

amirsaam:
10 month? nice.

amirsaam:
@Depal1 are you interested in this pr? if yes i shall fix its issues, otherwise close it pls.

Depal1:
> @Depal1 are you interested in this pr? if yes i shall fix its issues, otherwise close it pls.

I am, the UI changes sought with this PR would be very useful to manage large IPA collections for testing.

amirsaam:
> > @Depal1 are you interested in this pr? if yes i shall fix its issues, otherwise close it pls.
> 
> I am, the UI changes sought with this PR would be very useful to manage large IPA collections for testing.

Then I try to refactor and rebase it today.

PR #1087, opened 2023-09-08
Set Application Category Type

### LINKED ISSUE #1085: [Feature]: Game Mode support

No

It would be nice to have support for the new Game Mode feature in macOS Sonoma. As far as I am aware it turns on automatically and can't be manually enabled otherwise. I assume macOS auto-detects if a game is using Metal API and is set to launch in fullscreen, if these two conditions are met then it automatically turns on Game Mode which can be seen as a notification and icon in the top menu bar. 

iPhone and iPad games compatible on macOS can use Game Mode so I don't see why games launched from PlayCover can't also use it. 

image

### Anything else?

image

image

### Issue Language

TheMoonThatRises:
This allows the user to set the category of the app by adding the key `LSApplicationCategoryType` to the `Info.plist` located in the application. This can be changed under Misc in the application's settings.

More specifically, this feature enables and closes #1085 "Game Mode", which can be accomplished by the user setting the application type to `public.app-category.games` for those on MacOS Sonoma.

**This change is tested on MacOS Sonoma Developer Beta 23A5337a**. This should not affect the changes, but should still be tested on a stable version of MacOS.

Other Notes for reviewers:
- Should `LSApplicationCategoryType` be localised? Currently, the code is implemented but was too time consuming to implement it in the localisation files.
- Should the installer popup be updated to a `sheet` or something similar to accommodate for the user to specify the application category type on installation? The setting has been implemented to set all the installing types to the specified value but per installation type setting would be require some major refactoring.
- Should the application inherent the application type from the IPA? Since the application's `Info.plist` does not initially have the key `LSApplicationCategoryType`, it would have to be located elsewhere as iOS apps still have the application category types. I will look into that but since it is not already implemented, I am not too sure it will work.

### DIFF
--- a/PlayCover/Views/AppSettingsView.swift
+++ b/PlayCover/Views/AppSettingsView.swift
@@ -90,7 +90,8 @@ struct AppSettingsView: View {
                          closeView: $closeView,
                          hasPlayTools: $hasPlayTools,
                          hasAlias: $hasAlias,
-                         app: viewModel.app)
+                         app: viewModel.app,
+                         applicationCategoryType: viewModel.app.info.applicationCategoryType)
                     .tabItem {
                         Text("settings.tab.misc")
                     }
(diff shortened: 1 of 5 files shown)


JoseMoreville:
> Should LSApplicationCategoryType be localised? Currently, the code is implemented but was too time consuming to implement it in the localisation files.

Just add it to the english loc file


> Should the installer popup be updated to a sheet or something similar to accommodate for the user to specify the application category type on installation? The setting has been implemented to set all the installing types to the specified value but per installation type setting would be require some major refactoring

This can be discussed later, as you said, it'll require a major refactor of installer code

> Should the application inherent the application type from the IPA? Since the application's Info.plist does not initially have the key LSApplicationCategoryType, it would have to be located elsewhere as iOS apps still have the application category types. I will look into that but since it is not already implemented, I am not too sure it will work.

This is worth to take a look since it should be provided with the app plist but we can implement it as nil meanwhile

PlayCover/Views/AppSettingsView.swift
```diff
@@ -481,9 +482,34 @@ struct MiscView: View {
...
+                    Picker("", selection: $applicationCategoryType) {
+                        ForEach(LSApplicationCategoryType.allCases, id: \.rawValue) { value in
+                            Text(value.localizedName)
+                                .tag(value)
+                        }
+                    }
+                    .frame(width: 225)
+                    .onChange(of: applicationCategoryType) { _ in
+                        app.info.applicationCategoryType = applicationCategoryType
+                        Task(priority: .userInitiated) {
```
JoseMoreville, line 502: If possible this will be better as a function but this is fine

JoseMoreville (approved):

PR #639, opened 2022-12-27
Replace Usage of `DispatchQueue` with `Task`

IsaacMarovitz:

PR #1091, opened 2023-09-09
inject/remove PlayTools on other threads

aiden-leong:


TheMoonThatRises:
We've intentionally removed the usage of `DispatchQueue` (in #639). Is there another way of doing this? Also, when it is injecting PlayTools and resigning the app, it should be made where trying to modify the app or open the app is blocked.

TheMoonThatRises:
Closing in favour of changes presented in #1274.

PR #1116, opened 2023-10-02
Added M2 MacBooks with notch to notchModels

zanderp25:


TheMoonThatRises:
Instead of manually adding each MacBook version to this list (since a long list of new MacBook released this month), you could consider changing the notch detection method to the one discussed here: https://stackoverflow.com/questions/69685094/detect-macbook-notch-in-macos-monterey-and-higher. You'll have to test it though as I'm not sure if it will work or not.

zanderp25:
Thanks, I'll look into that

PR #1156, opened 2023-10-13
Added Support for iPhone 15 Pro Max in the iOS Device Picker

OZORDI:
Based on reputable sources like https://www.screensizes.app/?model=iphone-15-pro-max and https://everymac.com/systems/apple/iphone/specs/apple-iphone-15-pro-max-global-a3106-specs.html It is confirmed that the Device ID for the iPhone 15 Pro max is iPhone16,2.

Using this info i modified the AppSettings View file to include the device.


TheMoonThatRises:
Closing in favour of #1470.

PR #1176, opened 2023-10-24
IPA library app info

TheMoonThatRises:
This adds more transparency as to the information in each source, allowing the user to view the direct link (and access through the use of hyperlinks), checksum, and the other information provided by the source.
773a023d-c332-45c2-b69e-16fd1dc322ff

This will also show if the link to the app is valid and accessible, in the case of failed downloads.
Screenshot 2023-10-24 at 14 37 00

This is modelled after the settings view for apps the user already has installed, and can also be accessed through the "ⓘ" icon in the top navigation bar.
Screenshot 2023-10-24 at 14 38 30

### DIFF
--- a/PlayCover/en.lproj/Localizable.strings
+++ b/PlayCover/en.lproj/Localizable.strings
@@ -20,6 +20,18 @@
 "playapp.noSources.title" = "No IPAs Installed";
 "playapp.noSources.subtitle" = "You currently have no IPAs installed. Click the button below to import one.";
 
+"ipaLibrary.info" = "App Info";
+"ipaLibrary.info.title" = "%@ Info";
+"ipaLibrary.info.notAvailable" = "Link not valid for this app";
+"ipaLibrary.info.tab.info" = "Info";
+"ipaLibrary.info.appName" = "App Name:";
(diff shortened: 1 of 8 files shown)


PlayCover/en.lproj/Localizable.strings
```diff
@@ -20,6 +20,18 @@
 "playapp.noSources.title" = "No IPAs Installed";
 "playapp.noSources.subtitle" = "You currently have no IPAs installed. Click the button below to import one.";
 
+"ipaLibrary.info" = "App Info";
+"ipaLibrary.info.title" = "%@ Info";
+"ipaLibrary.info.notAvailable" = "Link not valid for this app";
+"ipaLibrary.info.tab.info" = "Info";
+"ipaLibrary.info.appName" = "App Name:";
+"ipaLibrary.info.bundleIdentifier" = "Bundle Identifier:";
```
Depal1: I suggest avoiding string duplicities in the localizations file by using some of those already defined for `settings.info.`.

Depal1 (changes requested):

Depal1 (approved):

PR #1130, opened 2023-10-08
Fix Discord RPC

TheMoonThatRises:
Fixes #1112 and requires https://github.com/PlayCover/PlayTools/pull/126, https://github.com/PlayCover/SwordRPC/pull/1. The fix is to symbolically link the discord-ipc socket "file" to the app's own temporary directory in `~/Library/Containers/[App ID]/Data/tmp` at the initialisation of each app model. In order to accomplish this, some of the functions and models have been changed, such as the `enumerate` function in FileManager and the `AppContainer` model itself. Along with that, the `PlayAppExtensions.swift` file was created to avoid going over the line limit for `PlayApp` with SwiftLint.

PR #1190, opened 2023-11-01
Add support for some iOS Apps that cannot find iOS Frameworks.

Ryu-ga:
On MacCatalyst, dyld load iOS frameworks linked already, but dynamically linking may not works due to difference between iOS filesystem and Mac.
iOS frameworks exist under /System/iOSSupport/System/Library/Frameworks.
Please add DYLD things to support such apps.

### DIFF
--- a/PlayCover/Model/PlayApp.swift
+++ b/PlayCover/Model/PlayApp.swift
@@ -8,7 +8,6 @@ import Foundation
 import IOKit.pwr_mgt
 
 class PlayApp: BaseApp {
-    private static let library = FileManager.default.homeDirectoryForCurrentUser.appendingPathComponent("Library")
     var displaySleepAssertionID: IOPMAssertionID?
     public var isStarting = false
 
@@ -16,6 +15,11 @@ class PlayApp: BaseApp {
         info.displayName.lowercased().appending(" ").appending(info.bundleName).lowercased()
     }
     var sessionDisableKeychain: Bool = false
+
+    override init(appUrl: URL) {
+        super.init(appUrl: appUrl)
+        self.loadDiscordIPC()
+    }
 
     func launch() async {
         do {
@@ -65,10 +69,6 @@ class PlayApp: BaseApp {
 
     func runAppExec() {
         let config = NSWorkspace.OpenConfiguration()
-
-        if settings.settings.injectIntrospection {
-            config.environment["DYLD_LIBRARY_PATH"] = "/usr/lib/system/introspection"
-        }
 
         NSWorkspace.shared.openApplication(
             at: url,
@@ -187,11 +187,11 @@ class PlayApp: BaseApp {
 
     lazy var playChainURL = PlayApp.playChainDirectory.appendingPathComponent(info.bundleIdentifier)
 
-    lazy var settings = AppSettings(info, container: container)
+    lazy var settings = AppSettings(info)
 
-    lazy var keymapping = Keymapping(info, container: container)
+    lazy var keymapping = Keymapping(info)
 
-    var container: AppContainer?
+    lazy var container = AppContainer(bundleId: info.bundleIdentifier)
 
     func hasPlayTools() -> Bool {
         do {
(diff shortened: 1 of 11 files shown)


Ryu-ga:
Some iOS app runtimes use dynamic linking.
https://github.com/robovm/robovm/blob/master/rt/robovm/src/main/java/org/robovm/rt/bro/Runtime.java
If a runtime use DYLD env variable to load frameworks, it cannot find them on MacCatalyst.
I guess there is need to set DYLD to pretend iOS environment for mac-catalyst unsupported apps.

ohaiibuzzle:
This *shouldn't* be an issue, you have an app that this particular patch would fix?
Because by default MacCatalyst apps should always load using `/System/iOSSupport`. Has been the case even on Intel Macs. Would be strange if Apple didn't support it in the Catalyst layer

Ryu-ga:
> This _shouldn't_ be an issue, you have an app that this particular patch would fix? Because by default MacCatalyst apps should always load using `/System/iOSSupport`. Has been the case even on Intel Macs. Would be strange if Apple didn't support it in the Catalyst layer

"Curse of Aros", The app actually doesn't catch dyld environment variables on Mac-catalyst.

2023-11-01 03:49:49.140009+0900 IOSLauncher[3583:64154] PC-DEBUG: PlayShadow is now loading
java.lang.UnsatisfiedLinkError: Library 'UIKit' not found
at org. robovm.rt. bro.Runtime.getHandle(Runtime.java :317)
at org. robovm. rt. bro. Runtime. loadLibrary (Runtime. java:191)
at org. robovm. rt.bro.Bro. bind (Bro.java: 60)
at org. robovm. objc.ObjCRuntime.bind(ObjCRuntime.java:92)
at org.robovm.apple.uikit.UIResponder.<clinit>(UIResponder.java:53)
at com.bitgate.curseofaros.IOSLauncher.main (IOSLauncher.java:67)

This is a log of the app, and it didn't crash but always quit due to DYLD env variables.

Pure iOS SwiftUI App
```
import SwiftUI

struct ContentView: View {
    var body: some View {
        Text("Test")
            .onAppear {
                if let dyld = ProcessInfo.processInfo.environment["DYLD_LIBRARY_PATH"] {
                    print("DYLD_LIBRARY_PATH env var: \(dyld)")
//DYLD_LIBRARY_PATH env var: /Users/[name]/Library/Developer/Xcode/DerivedData/Test-bfkvkgqricwfawdjxreioirkncqq/Build/Products/Debug-iphoneos:/usr/lib/system/introspection
                }
            }
    }
}
```
It print DYLD_LIBRARY_PATH like this. I guess Mac-Catalyst does not give DYLD env variables. This is fault of Mac-Catalyst but We can support such thing.

PlayCover/Model/PlayApp.swift
```diff
@@ -66,8 +66,10 @@ class PlayApp: BaseApp {
     func runAppExec() {
         let config = NSWorkspace.OpenConfiguration()
 
+        config.environment["DYLD_LIBRARY_PATH"] = "/System/iOSSupport/System/Library/Frameworks"
```
TheMoonThatRises: It would be best to have this be a setting under bypasses, as the environment should not be modified in the launch process. The if-statement below was a mistake when merging one of my PRs (and will be fixed here https://github.com/PlayCover/PlayCover/pull/1130/files#diff-8c70684d8da2457c4a34d4f5f9bf6124ac596cd5cdee1fc8a04a0ba2bef9d0f7L69).

Ryu-ga: I believe the problem is that MacCatalyst only supports linking with linked frameworks(I guess it is supported by dyld on Mac). In the iOS environment, iOS Frameworks are located in /System/Library/Frameworks. However, in MacCatalyst, that path is designated for MacOS, while iOS Frameworks exist in /System/iOSSupport/System/Library/Frameworks.
Since MacCatalyst does not support a "chroot" like functionality, it is unable to dynamically load System Plugin or Frameworks for iOS using the FileSystem. Linking via the filesystem is beneficial for certain non-objc or non-swift iOS apps, but due to the different file structure and lack of full chrooting in Mac, this may not be fully compatible. Therefore, I believe it should be the default, but I will follow your rules.

Ryu-ga: I fork your branch.
https://github.com/Ryu-ga/PlayCover/commit/25dcbac0d83ba0ae957cad8798c812d2dffcc23b,
It can be done while install phase.

PR #1195, opened 2023-11-05
Add support for some iOS Apps that cannot find iOS Frameworks.

Ryu-ga:
Add some apps that cannot find iOS Frameworks due to weird dynamic linking via FileSystem.

"Curse of Aros", The app actually doesn't catch dyld environment variables on Mac-catalyst.

2023-11-01 03:49:49.140009+0900 IOSLauncher[3583:64154] PC-DEBUG: PlayShadow is now loading
java.lang.UnsatisfiedLinkError: Library 'UIKit' not found
at org. robovm.rt.bro.Runtime.getHandle(Runtime.java :317)
at org. robovm.rt.bro.Runtime. loadLibrary (Runtime. java:191)
at org. robovm.rt.bro.Bro.bind (Bro.java: 60)
at org. robovm.objc.ObjCRuntime.bind(ObjCRuntime.java:92)
at org.robovm.apple.uikit.UIResponder.(UIResponder.java:53)
at com.bitgate.curseofaros.IOSLauncher.main (IOSLauncher.java:67)

This is a log of the app, and it didn't crash but always quit due to DYLD env variables.

Pure iOS SwiftUI App

import SwiftUI

struct ContentView: View {
    var body: some View {
        Text("Test")
            .onAppear {
                if let dyld = ProcessInfo.processInfo.environment["DYLD_LIBRARY_PATH"] {
                    print("DYLD_LIBRARY_PATH env var: \(dyld)")
//DYLD_LIBRARY_PATH env var: /Users/[name]/Library/Developer/Xcode/DerivedData/Test-bfkvkgqricwfawdjxreioirkncqq/Build/Products/Debug-iphoneos:/usr/lib/system/introspection
                }
            }
    }
}
It print DYLD_LIBRARY_PATH like this. I guess Mac-Catalyst does not give DYLD env variables.

### DIFF
--- a/PlayCover/Model/PlayApp.swift
+++ b/PlayCover/Model/PlayApp.swift
@@ -202,17 +202,18 @@ class PlayApp: BaseApp {
         }
     }
 
-    func introspection(set: Bool? = nil) -> Bool {
-        if info.lsEnvironment["DYLD_LIBRARY_PATH"] == nil {
-            info.lsEnvironment["DYLD_LIBRARY_PATH"] = ""
-        }
+    static let introspection: String = "/usr/lib/system/introspection"
+    static let iosFrameworks: String = "/System/iOSSupport/System/Library/Frameworks"
+
+    func changeDyldLibraryPath(set: Bool? = nil, path: String) -> Bool {
+        info.lsEnvironment["DYLD_LIBRARY_PATH"] = info.lsEnvironment["DYLD_LIBRARY_PATH"] ?? ""
 
         if let set = set {
             if set {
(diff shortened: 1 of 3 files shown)


JoseMoreville:
I haven't test this yet but wouldn't this cause side effects?

Ryu-ga:
> I haven't test this yet but wouldn't this cause side effects?

I don't think there are any side effects because this is only an option for very specific games (robovm based apps). However, Swift-Lint seems to be going to argue that the class code in PlayApp.swift is too long when adding more code. This code does not hold much value. If this pr was closed due to mess code, I would create forked app for this issue and provide zipped app similar to `Sonoma Quick Patch` of mine.

JoseMoreville (approved):

PR #1207, opened 2023-11-10
Added store app icon load timeout

Wind-Explorer:
loads for 6 seconds

then times out to indicate failure to load


PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -85,6 +86,12 @@ struct StoreAppConditionalView: View {
 
     @State private var cache = DataCache.instance
 
+    func waitForIconLoad() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 6) {
```
TheMoonThatRises: This should be switched out to Swift 5.5's concurrency async and await instead of DispatchQueue. You can see an example of how to do a delay after x seconds here: https://stackoverflow.com/questions/59682446/how-to-trigger-an-action-after-x-seconds-in-swiftui

TheMoonThatRises (changes requested):

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -85,6 +86,11 @@ struct StoreAppConditionalView: View {
 
     @State private var cache = DataCache.instance
 
+    @Sendable private func waitForIconLoad() async {
+        try? await Task.sleep(nanoseconds: 6_000_000_000)
```
TheMoonThatRises, line 91: `loadingLocalIcon` should be set to `true` at the start of the function (before the sleep), as the image might reload (such as during grid view) but would have the error icon instead of the loading one.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -154,9 +166,16 @@ struct StoreAppConditionalView: View {
...
                                          .fill(.regularMaterial)
                                          .overlay {
-                                             ProgressView()
-                                                 .progressViewStyle(.circular)
+                                             if loadingLocalIcon {
+                                                 ProgressView()
+                                                     .progressViewStyle(.circular)
+                                             } else {
+                                                 Image(systemName: "exclamationmark.triangle")
+                                                     .font(.system(size: 24))
```
TheMoonThatRises, line 175: Setting the font isn't needed.

TheMoonThatRises (changes requested):

PlayCover/Views/App Views/StoreAppView.swift
Wind-Explorer: > Setting the font isn't needed.

The purpose of setting a font is to specify a size for the icon, to ensure that the icon is of a reasonable size to the space available. Without specifying a size, the icon may appear too small to view, like so:


截屏2024-06-26 上午11 35 13


whereas with a font size specified:
截屏2024-06-26 上午11 36 33


especially considering how the spinner icon is already of a relatively large size, it would only make sense that they match.
截屏2024-06-26 上午11 37 03

Let me know what you think!

TheMoonThatRises (approved):

PR #1208, opened 2023-11-10
Update Localizable.strings for correcting terminology and grammars  in Traditional Chinese.

vincent5753:


XuYicong:
All translations are handled via [Weblate](https://hosted.weblate.org/projects/playcover/). Please submit translations there instead of opening PRs.

PR #1274, opened 2023-12-15
Adopt asynchronous methods to avoid UI blocking

Stv-X:
These changes avoid the stuck with macOS spinning beach ball caused by UI blocking while performing long-consuming work.

before:
before

after:
after


TheMoonThatRises:
Overall, this PR looks good. But could you add the same thing to other UI blocking elements (they should be the ones that also resign the app)?

Stv-X:
Thanks for the review! I'm working on adding the same content to other UI elements.

Stv-X:
@TheMoonThatRises Hi, it seems there are no UI elements left behind any more. I believe this PR is ready to be merged.

PlayCover.xcodeproj/project.pbxproj
```diff
@@ -72,6 +72,8 @@
 		AA71964B287A0EA800623C15 /* PlayRules.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA71964A287A0EA800623C15 /* PlayRules.swift */; };
 		AA71964D287A35F400623C15 /* com.nexon.bluearchive.yaml in Resources */ = {isa = PBXBuildFile; fileRef = AA71964C287A35F400623C15 /* com.nexon.bluearchive.yaml */; };
 		AA818CB5287ABEC3000BEE9D /* Yams in Frameworks */ = {isa = PBXBuildFile; productRef = AA818CB4287ABEC3000BEE9D /* Yams */; };
+		AACF11A42B31769C008BAA81 /* UnintallVM.swift in Sources */ = {isa = PBXBuildFile; fileRef = AACF11A32B31769C008BAA81 /* UnintallVM.swift */; };
```
TheMoonThatRises: Typo in file name (`UnintallVM.swift` -> `UninstallVM.swift`).

PlayCover/Views/AppSettingsView.swift
```diff
@@ -21,6 +21,10 @@ struct AppSettingsView: View {
     @State var hasPlayTools: Bool?
     @State var hasAlias: Bool?
 
+    @State private var isSwitchingPlayToolsInjection = false
```
TheMoonThatRises: Is it possible to combine these into one variable (and it's references below)? Maybe something like `isInBlockingTask`.

Stv-X: I think it is necessary for users to distinguish which task is in progress, which depends on using independent variables for control. 
Is it ok to use enumerate type as an alternative to simplify code?
Like:
```swift
enum BlockingTask {
    case none, playTools, introspection, iosFrameworks
}
```
```swift
struct AppSettingsView: View {

//    @State private var isSwitchingPlayToolsInjection = false
//    @State private var isSwitchingIntrospection = false
//    @State private var isSwitchingIosFrameworks = false
    @State private var currentTask = BlockingTask.none
    private var isInBlockingTask: Bool { currentTask != .none }
    
    var body: some View {
        VStack {
            ...
            ByPassesView(settings: $viewModel.settings,
                                      hasPlayTools: $hasPlayTools,
//                                      isSwitchingIntrospectionLibraryInsertion: $isSwitchingIntrospection,
//                                      isSwitchingIosFrameworksInsertion: $isSwitchingIosFrameworks,
                                      task: $currentTask,
                                      app: viewModel.app)
            ...
        }
        .disabled(isInBlockingTask)
        ...
    }
}
```
```swift
struct BypassesView: View {
//    @Binding var isSwitchingIntrospectionLibraryInsertion: Bool
//    @Binding var isSwitchingIosFrameworksInsertion: Bool
    @Binding var task: BlockingTask

    Toggle("settings.toggle.introspection", isOn: $hasIntrospection)
     .toggleStyle(.async($task, role: .introspection))
     .onChange(of: hasIntrospection) {_ in
          task = .introspection
          Task {
               _ = await app.changeDyldLibraryPath(set: hasIntrospection, path: PlayApp.introspection)
               task = .none
          }
     }

}
```

TheMoonThatRises: Yeah, that would be good.

Stv-X: Done.

PlayCover/Views/App Views/ClearCacheModalView.swift
```diff
@@ -0,0 +1,84 @@
...
+            .frame(width: 60, height: 60)
+            .padding()
+            Text("alert.app.delete")
+                .bold()
+                .frame(width: 250)
+        }
+        .padding()
+        .fixedSize()
+        .toolbar {
+            ToolbarItem(placement: .destructiveAction) {
```
TheMoonThatRises: The proceed button should be on the right side.

PlayCover/Views/Uninstaller.swift
```diff
@@ -101,18 +101,19 @@ class Uninstaller {
...
-            uninstall(app)
+            Task { await uninstall(app) }
         }
     }
 
-    static func uninstall(_ app: PlayApp) {
+    static func uninstall(_ app: PlayApp) async {
         var uninstallNum = 0
 
+        UninstallVM.shared.next(.begin, 0.0, 0.0)
```
TheMoonThatRises: There should be stages to the uninstall progress bar, similar to the other progress vms for a better user experience. You could also consider adding the uninstall vm to `clearAllCache` instead, as that is the one that consumes the most time, and progress could be easily tracked as it goes through a set amount of directories.

PlayCover/en.lproj/Localizable.strings
```diff
@@ -90,6 +90,8 @@
 "playapp.install.signing" = "Signing app...";
 "playapp.install.addToLib" = "Adding app to library...";
 "playapp.install.copy" = "Copying app...";
+"playapp.uninstall.begin" = "Uninstalling app...";
+"playapp.uninstall.finished" = "Finished";
```
TheMoonThatRises: It would be better to use `playapp.progress.finished` as it is the same thing and more generic for progresses.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -477,22 +484,36 @@ struct BypassesView: View {
...
         }
         .onChange(of: hasIosFrameworks) {_ in
-            _ = app.changeDyldLibraryPath(set: hasIosFrameworks, path: PlayApp.iosFrameworks)
+            task = .iosFrameworks
+            Task {
+                _ = await app.changeDyldLibraryPath(set: hasIosFrameworks, path: PlayApp.iosFrameworks)
+                task = .none
+            }
+        }
+        .task {
```
TheMoonThatRises: Loading these variables should be kept passed in, as this is not a blocking task. Another thing with having the variable being set here is that it triggers the `onChange` event, which causes the loading to take unnecessarily longer.

PlayCover/Views/App Views/ClearCacheModalView.swift
```diff
@@ -0,0 +1,84 @@
...
+                if isClearingCache {
```
TheMoonThatRises: This `if-else` statement is redundant, as the opacity is changed and has a `ProgressView` overlayed on the button text below.

PlayCover/Views/App Views/PlayAppView.swift
```diff
@@ -136,13 +139,15 @@ struct PlayAppView: View {
...
-            }
+
+        /*
+         * Since SwiftUI‘s native alert does not support being dismissed programmatically,
+         * an async method will immediately hide the dialog after clicking the button,
+         * this leads to the execution state of the task not being observed by the user,
+         * resulting in the risk of misoperation,
+         * so a customized sheet is used as an alternative.
+         */
+            .sheet(isPresented: $showClearCacheAlert) { ClearCacheModalView(app: $app) }
```
TheMoonThatRises: The uninstaller and the cache clearer should be consistent in alert and progress view. It would be best to change this to the uninstaller format (regular alert with progress vm showing uninstall progress), or changing the uninstaller alert to match this. The `ClearCacheModelView` could be generalised to support both of these processes.

PlayCover.xcodeproj/project.pbxproj
```diff
@@ -72,6 +72,8 @@
 		AA71964B287A0EA800623C15 /* PlayRules.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA71964A287A0EA800623C15 /* PlayRules.swift */; };
 		AA71964D287A35F400623C15 /* com.nexon.bluearchive.yaml in Resources */ = {isa = PBXBuildFile; fileRef = AA71964C287A35F400623C15 /* com.nexon.bluearchive.yaml */; };
 		AA818CB5287ABEC3000BEE9D /* Yams in Frameworks */ = {isa = PBXBuildFile; productRef = AA818CB4287ABEC3000BEE9D /* Yams */; };
+		AACF11A42B31769C008BAA81 /* UninstallVM.swift in Sources */ = {isa = PBXBuildFile; fileRef = AACF11A32B31769C008BAA81 /* UninstallVM.swift */; };
+		AACF11A62B3176B2008BAA81 /* ClearCacheModalView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AACF11A52B3176B2008BAA81 /* ClearCacheModalView.swift */; };
```
TheMoonThatRises: Typo, should be `ClearCacheModelView` (change "a" in "Modal" to "e"). The struct name has the same typo.

TheMoonThatRises (changes requested):
Overall, pretty good work. Just needs some changes here and there.

PlayCover/en.lproj/Localizable.strings
Stv-X: I think it is necessary to retain `playapp.uninstall.finished` for distinction purposes. In the localization of other languages, the translated content might include contextual information based on the language's usage habits to provide a more comfortable reading experience.

For instance, in a Simplified Chinese environment, `playapp.progress.finished` is translated as "安装完成" which should literally be understood as "Installation Complete" rather than the English version's "Finished." This results in an inappropriate display for users if `playapp.progress.finished` is retained, as they would see "Installation Complete" after uninstallation, which is evidently incorrect.

PlayCover.xcodeproj/project.pbxproj
Stv-X: "Modal" is not a typo, which refers to a dialog box like an alert, popover, and sheet. 
Check this out: [https://developer.apple.com/documentation/swiftui/modal-presentations#Overview](url)

PlayCover/Views/Uninstaller.swift
Stv-X: Based on my observation, it seems that 'clearAppData' is the only task that consumes time. Other tasks are completed instantly, and even if suitable stage values are set for each tasks, it will not make a difference to the visual presentation of the UI at all.💦

However, I am currently attempting to modify the interface style for clearAllCache to match the native alert with progress bar style used by uninstaller. This is a good idea.

TheMoonThatRises:
Sorry it's been a bit since the last review, but I'll take a look at this again after I try and fix a bug that this PR is impacted by.

PlayCover/ViewModel/UninstallVM.swift
```diff
@@ -0,0 +1,22 @@
+//
```
TheMoonThatRises: I don't think the uninstaller needs the non-blocking function, as the cache clearing function is now pretty fast due to a bug fix (<2 seconds for Genshin).

TheMoonThatRises (changes requested):

PlayCover/ViewModel/UninstallVM.swift
Stv-X: Got it, UninstallVM and related code have been removed in the latest commit.

TheMoonThatRises (approved):

TheMoonThatRises:
Before I merge, theres actually one last blocking action in settings that was added somewhat recently. Can you see if its possible to do the loading for changing application type (under misc)?

Stv-X:
Hi, @TheMoonThatRises ,

I’ve addressed the blocking action you mentioned in the MiscView related to the application type change by making the Shell.signApp method asynchronous. 

This PR is ready to be merged.

PlayCover/Utils/PlayTools.swift
```diff
@@ -83,7 +83,7 @@ class PlayTools {
             if result {
                 do {
                     try installPluginInIPA(exec.deletingLastPathComponent())
-                    try Shell.signApp(exec)
+                    Task { try await Shell.signApp(exec) }
```
TheMoonThatRises: This app sign and the one in `removeFromApp` does not force the user to wait for the app to be fully signed, which causes some issues when opening the app without it being fully signed.

TheMoonThatRises (changes requested):

PlayCover/Utils/PlayTools.swift
Stv-X: Got it, this time I used GCD to deal with this kind of code that needs to be executed synchronously on some occasions.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -523,12 +560,16 @@ struct MiscView: View {
...
                     .frame(width: 225)
                     .onChange(of: applicationCategoryType) { _ in
+                        task = .applicationCategoryType
                         app.info.applicationCategoryType = applicationCategoryType
                         Task(priority: .userInitiated) {
-                            do {
-                                try Shell.signApp(app.executable)
-                            } catch {
-                                Log.shared.error(error)
+                            DispatchQueue.global(qos: .background).async {
```
TheMoonThatRises: This `DispatchQueue` inside the `Task` isn't necessary.

TheMoonThatRises (changes requested):

PlayCover/Views/AppSettingsView.swift
Stv-X: According to my observation, when only `Task` is used, UI will respond to all operations performed by the user during the waiting period after the end of the blocking, which may cause some functions be accidentally triggered, so I chose to use `DispatchQueue.global(qos: .background)`. In the latest commit, I removed the `Task` in this code to ensure simplicity.

TheMoonThatRises: What UI operations? I can't seem to replicate this.

Stv-X: I recorded two videos to compare and show the differences in the results of these two codes with only the following differences (I wrote a demo program to show the problems more clearly, and they behave exactly the same in PlayCover)
`Task(priority: .userInitiated) { doSomething() }` 
https://github.com/PlayCover/PlayCover/assets/30586070/9c579568-9a92-4e89-b806-575366b69b13

`DispatchQueue.global(qos: .background).async { doSomething() }`
https://github.com/PlayCover/PlayCover/assets/30586070/c21ed8b5-25c8-4116-9a61-c5179ccf80c2

TheMoonThatRises: Try `Task.detached` instead of `Task(priority: .userInitiated)`.

Stv-X: Wow, that looks better. I have added this change to the latest commit.

TheMoonThatRises (approved):

PR #1282, opened 2023-12-19
fix: missing com.apple.developer.applesignin entitlement

devnoname120:
Note that I couldn't try this fix because I don't have an Apple Developer License so I can't sign PlayCover after building it.
Reference: https://stackoverflow.com/a/63804071/3634271

**How to reproduce**:

1) Download [`ch.poinz.Poinz_4.1.14_und3fined.ipa`](https://web.archive.org/web/20231219082608/https://36.filelu.com/d/rr3nw7tjjqqnjjtadcm44z53thpxsukqdu2ksshbu3vs7fca2fti74ovm5wbljwzexwzzbp6/ch.poinz.Poinz_4.1.14_und3fined.ipa) and install it in PlayCover.

2) Click on sign-in with Apple:
image

3) The following error appears:
image

### DIFF
--- a/PlayCover/Utils/Entitlements.swift
+++ b/PlayCover/Utils/Entitlements.swift
@@ -51,6 +51,7 @@ class Entitlements {
         base["com.apple.security.personal-information.calendars"] = true
         base["com.apple.security.personal-information.location"] = true
         base["com.apple.security.print"] = true
+        base["com.apple.developer.applesignin"] = "Default"
     }
 
     // swiftlint:disable:next cyclomatic_complexity


ohaiibuzzle:
I doubt this would work, Keep in mind that apps signed by PlayCover do not have proper Apple Developer Codesign either, they are localsigned on your device only, with a bogus TeamID to boot.
So this would only work if fakesigned iOS apps can sign in with Apple IDs, which I believe they can't
Signing PlayCover itself doesn't enable Apple ID sign in, it's done merely so we don't get the nasty "Unknown App" popup after downloading the app and confuse users.

devnoname120:
@ohaiibuzzle Hmmm... is there a way to test this? Alternatively, could there be a way to inject the app into the executable memory of one that has the right entitlements? Or a fake implementation like PlayChain?

devnoname120:
@ohaiibuzzle Hmmm... is there a way to test this? Alternatively, could there be a way to inject the app into the executable memory of one that has the right entitlements? Or a fake implementation like PlayChain?

ohaiibuzzle:
@devnoname120 Yes actually. Simply force the app to re-sign (eg. reinstall with your modifications on) and then try to sign in with an Apple ID. If it succeed, your changes work, if it doesn’t… well...

ohaiibuzzle:
@devnoname120 PlayChain works because keychain ops are local, which we can simply spoof with our implementation that doesn’t care about signatures (and it is not even good at that in many aspects, I really want to rethink PlayChain at some point).

Apple ID sign in is a handshake between your app and Apple, very slim chances for things like that to happen

PR #1472, opened 2024-05-16
Google drive hosted IPA

Catta1997:
I noticed some IPA (like HSR or Genshin) on `Decrypt.day` are hosted in Google Drive, this PR add a redirect catcher and a scraping to detect all I need to create a direct link to the IPA to bypass "Antivirus error" that show on too big file.

Steps that I need to do
- [x] Redirect catch 
- [x] Scrape and obtain new direct link
- [x] Pass the new link to the native download function
- [x] Pass the new url before the native-download function so don't need to change `url` variable from `let` to `var`
- [x] Find a correct way to put the native-download function in sleep until redirect and scraping are completed


Catta1997:
image
it works as expected, sometimes it not work for Google Drive quota reached, cannot bypass it without a login (maybe I can work on it but not sure if is ok)

Catta1997:
I have to understand how to correctly import SwiftSoup, any help appreciated

Catta1997:
I managed to install SwiftSoup with Carthage but I'm not sure if I did it correctly

PlayCover/Utils/GoogleDrive.swift
```diff
@@ -0,0 +1,126 @@
...
+            if error != nil {
+                return
+            }
+        }
+        task.resume()
+    }
+    func scrapeWebsite(from request: URLRequest) {
+        dispatchGroup.enter() // Enter group
+        let task = session.dataTask(with: request) { data, _, error in
+            defer { self.dispatchGroup.leave() } // Esci dal gruppo al termine della task
```
XuYicong: Please avoid non-English comment

PlayCover.xcodeproj/project.pbxproj
```diff
@@ -83,6 +83,9 @@
 		B1419FB628BA82EE000CB69F /* DiscordActivity.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1419FB528BA82EE000CB69F /* DiscordActivity.swift */; };
 		B17FD03C28C70C2100B1D4CA /* CoreUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B17FD03828C70BAE00B1D4CA /* CoreUI.framework */; };
 		B17FD04728C7B0D900B1D4CA /* AssetsExtractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = B17FD04628C7B0D900B1D4CA /* AssetsExtractor.swift */; };
+		B4699A482BF60F1A00D4B521 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
XuYicong: Filename is null😭 better fix this if you can

Cartfile
```diff
@@ -1 +1,2 @@
 github "PlayCover/PlayTools" "master"
+github "scinfu/SwiftSoup"
```
XuYicong: This doesn't look like the right way to import a package? I mean there are other imported packages like `inject` and `Sparkle` but they're not written here. I think they're configured in the project configuration? I'm not sure what the difference is tho.

PlayCover/Views/App Views/StoreAppView.swift
```diff
@@ -37,8 +37,16 @@ struct StoreAppView: View {
                 if downloadVM.inProgress {
                     Log.shared.error(PlayCoverError.waitDownload)
                 } else {
-                    DownloadApp(url: url, app: app,
-                                warning: warningMessage).start()
+                    let redirectHandler = RedirectHandler() // checking page redirect
+                    redirectHandler.redirectCatch(from: url)
+                    redirectHandler.waitForAllTasksToComplete()
```
XuYicong: I think it's better to hide these implementation details (`redirectCatch` and `waitForAllTasksToComplete`) inside the `RedirectHandler` class. And the `finalURL` should have something like public getter and private setter? 

I prefer styles like
```
let landingUrl = RedirectHandler.catch(from: url)
```

And also hide the check for `"drive"` inside the class. i.e. return the original URL if there are no redirections.

Cartfile
Depal1: Indeed, the dependency should be added through the Swift Package Manager. That way we keep Carthage exclusively for PlayTools. https://developer.apple.com/documentation/xcode/adding-package-dependencies-to-your-app

PlayCover/Views/App Views/StoreAppView.swift
Catta1997: Done, now the variable finalURL is private with getter and setter, initialized during class construction with the "reale" url so the "drive" check is no more necessary since if finalURL is not edited getter will return the "real" one

PlayCover/Utils/GoogleDrive.swift
Catta1997: Translated

Cartfile
Catta1997: Will try to do it, I already tried add it with package manager (same as Sparkle)  but build fail. Will try to fix it

PlayCover.xcodeproj/project.pbxproj
Catta1997: fixed

Cartfile
Catta1997: ok, I think I did it but my Carthage is bugged and I skipped the `Carthage Bootstrap` build phase but I did it with terminal. I think I have Cache somewhere but I removed all path I found

Catta1997: I fixed this changing SwiftSoup version

Depal1 (approved):

PR #1511, opened 2024-06-11
Fix Settings render issues

Catta1997:
Start working on fix render issue of Settings UI where the toggle not update (state change but not graphically) 

- [x] KeyMapView
- [x] GraphicView
- [x] BypassView
- [x] MiscView
- [x] Check if ALL toggle are saved correctly

### DIFF
--- a/PlayCover/Views/AppSettingsView.swift
+++ b/PlayCover/Views/AppSettingsView.swift
@@ -152,20 +152,37 @@ struct AppSettingsView: View {
 
 struct KeymappingView: View {
     @Binding var settings: AppSettings
-
+    @AppStorage("settings.settings.keymapping") private var keymapping = false
+    @AppStorage("settings.settings.noKMOnInput") private var noKMOnInput = false
+    @AppStorage("settings.settings.enableScrollWheel") private var enableScrollWheel = false
+    init(settings: Binding<AppSettings>){
+        self._settings = settings
+        self.keymapping = settings.settings.keymapping.wrappedValue
+        self.noKMOnInput = settings.settings.noKMOnInput.wrappedValue
+        self.enableScrollWheel = settings.settings.enableScrollWheel.wrappedValue
+    }
     var body: some View {
         ScrollView {
             VStack {
                 HStack {
                     Toggle("settings.toggle.km", isOn: $settings.settings.keymapping)
                         .help("settings.toggle.km.help")
+                        .onChange(of: settings.settings.keymapping) { _ in
+                            settings.settings.keymapping = settings.settings.keymapping
+                        }
                     Spacer()
                     Toggle("settings.toggle.autoKM", isOn: $settings.settings.noKMOnInput)
                         .help("settings.toggle.autoKM.help")
+                        .onChange(of: settings.settings.noKMOnInput) { _ in
+                            settings.settings.noKMOnInput = settings.settings.noKMOnInput
+                        }
                 }
                 HStack {
                     Toggle("settings.toggle.enableScrollWheel", isOn: $settings.settings.enableScrollWheel)
                         .help("settings.toggle.enableScrollWheel.help")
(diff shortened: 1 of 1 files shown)


Catta1997:
https://github.com/PlayCover/PlayCover/assets/34333654/b516a846-a77d-40e3-a70a-2e402623fbdb

Catta1997:
Should be all, not totally sure if all settings is saved correctly. I checked MetalHUD and DiscordRP and both are correct

mitchellmebane:
I built this branch and all the controls I tested work properly.

Clicking the "Fix window display issues" checkbox on the Graphics tab doesn't enable the dropdown next to it, but I've never used that, so I don't know how it's supposed to work. I can't seem to enable that dropdown in the old nightly 636, either.

mitchellmebane:
After looking through the code, I see that the dropdown only gets enabled when "Fix window display issues" is checked _and_ "Resolution" is set to "App Default".  It's working properly in this branch.  👍

TheMoonThatRises:
I think we should wait for an update from Apple, as this is most likely a bug and not intended behaviour. This does not occur on previous versions.

Catta1997:
> I think we should wait for an update from Apple, as this is most likely a bug and not intended behaviour. This does not occur on previous versions.

oh ok, btw this is a thing in other part of the project ([Settings/InstallSettings.swift](https://github.com/PlayCover/PlayCover/blob/a14960e031fb703ca0e07a1d376a52623db5e0f6/PlayCover/Views/Settings/InstallSettings.swift#L13), [PlayCover/Views/Settings/UpdateSettings.swift](https://github.com/PlayCover/PlayCover/blob/a14960e031fb703ca0e07a1d376a52623db5e0f6/PlayCover/Views/Settings/UpdateSettings.swift#L13) and [PlayCover/Views/Settings/UninstallSettings.swift](https://github.com/PlayCover/PlayCover/blob/a14960e031fb703ca0e07a1d376a52623db5e0f6/PlayCover/Views/Settings/UninstallSettings.swift#L13)).

mitchellmebane:
I may be way off base here, but is the issue the multi-level object nesting?  E.g., in the `AppSettingsView` tab views, [the views are binding to an instance of `AppSettings`](https://github.com/PlayCover/PlayCover/blob/3.0.0/PlayCover/Views/AppSettingsView.swift#L188), but [the controls are using an attribute of the nested `AppSettingsData`](https://github.com/PlayCover/PlayCover/blob/3.0.0/PlayCover/Views/AppSettingsView.swift#L217).

> ```swift
> Picker("", selection: $settings.settings.iosDeviceModel)
> ```

All the docs and examples of SwiftUI data binding I've seen either use a directly published property, or use a property of a published class. I haven't seen anything which uses a property in a struct nested in a published class.

TheMoonThatRises:
> I may be way off base here, but is the issue the multi-level object nesting? E.g., in the `AppSettingsView` tab views, [the views are binding to an instance of `AppSettings`](https://github.com/PlayCover/PlayCover/blob/3.0.0/PlayCover/Views/AppSettingsView.swift#L188), but [the controls are using an attribute of the nested `AppSettingsData`](https://github.com/PlayCover/PlayCover/blob/3.0.0/PlayCover/Views/AppSettingsView.swift#L217).
> 
> > ```swift
> > Picker("", selection: $settings.settings.iosDeviceModel)
> > ```
> 
> All the docs and examples of SwiftUI data binding I've seen either use a directly published property, or use a property of a published class. I haven't seen anything which uses a property in a struct nested in a published class.

I believe that this is the issue. @Catta1997 if you want to try fixing it by removing the nesting of classes and structures, do try. If not, I can merge this as a temporary fix.

Catta1997:
Ok, I will try

Catta1997:
But since I don know a lot SwiftUI I think you can merge this temporary fix 'cause I don't know if I will be able to fix it in the other way

TheMoonThatRises (approved):

PR #1517, opened 2024-06-16
Auto re-sign app on launch

XuYicong:
When running `codesign -dv` on some game, its signature info would be displayed. If the `Info.plist` has bad signature, a line would be `Info.plist=not bound`. If it's good, the line would be like `Info.plist entries=56`. In this way we can quickly check for a signature failure that would lead to an `OSStatus-67030` error and cause a crash. So we can re-sign the app if such case is detected.

The shell command part is working perfectly, but I have some problems regarding SwiftUI and need help.

Now that it would auto re-sign app on launch, a launch could take a long time. Therefore, a progress view is needed to give the user some feedback. 
截屏2024-06-17 03 40 04
截屏2024-06-17 03 40 23

This works perfectly with a `showStartingProgress` variable, but when `isList` is changed during starting, `showStartingProgress` is reset to `false`.

I then decide to rely on `app.isStarting` and update `showStartingProgress` from it, but something I don't understand happens: `app.isStarting` also resets to `false`. This not only lead to the not showing of the progress view when switching between List view and Grid view, but also makes the app can be launched multiple times during launching. This seems an existing issue, but it was not severe because launch time was short. But now it becomes severe, I don't know why it would reset and don't know how to fix it. 

Edit: Thanks to @Wind-Explorer the above issue is fixed.


Depal1:
@Wind-Explorer It seems like there are non committed changes since `PlayAppVM` is not in scope.

TheMoonThatRises:
I've tested this, but only with already signed applications. Is there a way to reliably test this method against applications with invalid signatures?

Depal1:
> I've tested this, but only with already signed applications. Is there a way to reliably test this method against applications with invalid signatures?

Modify any parameter in Info.plist, something noticeable like `CFBundleName` provides visual input too.

Depal1 (approved):

TheMoonThatRises (approved):

PR #1518, opened 2024-06-19
Update check

Catta1997:
I thought to add an option to the contextual menu of an app to check for update.
When you click on "Check for update" PlayCover will search the app name on the library and then will compare the BundleID to find the correct one. When (and if) finded will compare the 2 version to detect if is new or not

- [x] Use the same function of "StoreAppView.swift" instead rewrite it
- [ ] Check for update on start (if not too "expensive")
- [ ] New tab (under IPA Library) with list of all app that has new version, double click -> update (no idea on how to do it)

Checking on startup and a tab are "expensive" since I have to check the update for each app and the only way I've found is to simulate a search, check the bundle ID (100% sure the app is the corrrect one) and check the version. I think that the best way to add it is to right click -> check update


Catta1997:
https://github.com/PlayCover/PlayCover/assets/34333654/ee9bab8f-395a-4ec4-9e5b-b5a9b2bf5e5c

PR #1543, opened 2024-07-02
Check if app has macOS native version

Catta1997:


PlayCover/Model/PlayApp.swift
```diff
@@ -32,12 +33,72 @@ class PlayApp: BaseApp {
     }
     var sessionDisableKeychain: Bool = false
 
+    func fetchAppID(bundleID: String, completion: @escaping (String?) -> Void) {
```
amirsaam: we already have an itunes lookup function, use that.
`PlayCover/Model/ITunesResponse.swift`

Catta1997: ohhhh, tysm

PlayCover/Model/PlayApp.swift
```diff
@@ -33,53 +33,35 @@ class PlayApp: BaseApp {
...
-                    completion(nil)
-                }
-            } catch {
-                completion(nil)
-            }
+        let itunes: ITunesResponse? = await getITunesData(urlString)
+        if let appID = itunes?.results.first?.trackId {
+            return appID
+        } else {
+            return 0
```
amirsaam:
```swift
guard let appID = itunes?.results.first?.trackId else { return 0 }
return appID
```

Catta1997: done, thanks

PlayCover/en.lproj/Localizable.strings
```diff
@@ -307,3 +307,8 @@
 
 "keycover.alert.title" = "Keychain was not unlocked";
 "keycover.alert.content" = "Keychain access has been disabled for the current session";
+
+"macos.version" = "This app has a native MacOS version on AppStre";
```
amirsaam: `This app can be installed from Mac App Store`

PlayCover/en.lproj/Localizable.strings
```diff
@@ -307,3 +307,8 @@
...
+"alert.start.anyway" = "Start Anyway";
+"alert.quit" = "Quit";
```
amirsaam: use the cancel instead, it does have a key already

PlayCover/en.lproj/Localizable.strings
```diff
@@ -307,3 +307,8 @@
...
+"alert.open.appstore" = "Open AppStore";
```
amirsaam: `Open Mac App Store`

PlayCover/Model/PlayApp.swift
```diff
@@ -38,6 +39,10 @@ class PlayApp: BaseApp {
             if prohibitedToPlay {
                 await clearAllCache()
                 throw PlayCoverError.appProhibited
+            } else if hasMacVersion, await !runMacOSWarning() {
+                await clearAllCache()
```
amirsaam: you don't need to clear the cache here, it will remove user data if they just wanted to check it your provided data out

amirsaam:
also, make playcover remember user choice for that app and don't bother them to go through the alert every time

Catta1997:
> also, make playcover remember user choice for that app and don't bother them to go through the alert every time

ok, i will look how to do it

amirsaam:
LGTM

PlayCover/Model/PlayApp.swift
```diff
@@ -6,12 +6,12 @@
 import Cocoa
 import Foundation
 import IOKit.pwr_mgt
+import StoreKit
```
TheMoonThatRises: Does `StoreKit` need to be imported here?

PlayCover/Model/PlayApp.swift
```diff
@@ -35,9 +35,13 @@ class PlayApp: BaseApp {
     func launch() async {
         do {
             isStarting = true
+            let noMacAlert = UserDefaults.standard.bool(forKey: "\(info.bundleIdentifier).noMacAlert")
             if prohibitedToPlay {
                 await clearAllCache()
                 throw PlayCoverError.appProhibited
+            } else if hasMacVersion && !noMacAlert, await !runMacOSWarning() {
```
TheMoonThatRises: I think it would be best to have this alert ran when the user wants to install the application instead of launching it.

Catta1997: in my opinion it is better at startup because it works if you already have an app installed or if an app will be added to the list. but if you want I'll change it or add both, since app will remind the settings

TheMoonThatRises:
I'll check this out and merge it soon. I've been pretty busy lately (give me 1~3 days).

TheMoonThatRises:
When clicking the `Open in App Store` button, PlayCover crashes in the `getITunesData` function. This could be an issue with the MacOS beta, can someone else check this?

PlayCover/Model/PlayApp.swift
Depal1: I agree with Moon, I would make it install time only.

TheMoonThatRises: Currently, it seems that you added it to the Downloader instead of the Installer, which gives the incorrect behaviour.

Depal1 (approved):

Julmik647:
Does it work only for App Store apps? There are many apps that have Mac version, but it's not on App Store.

Catta1997:
> Does it work only for App Store apps? There are many apps that have Mac version, but it's not on App Store.

yes only app store, no way to automatically detect apps that have macos version not on app store

PR #1592, opened 2024-07-16
Feat: SourceFolders

amirsaam:
Originally #880, messed up the codes so I needed to detach my last commit to resolve conflicts again.


TheMoonThatRises:
Can this be implemented without the Semaphore and NavigationStack-backport packages? It would be nice to not include too many external packages, as most of this can be done in pure Swift (as for NavigationStack, I was thinking it could be a feature restricted to the earliest MacOS version that supports it). @Depal1?

amirsaam:
@TheMoonThatRises

`NavigationStack` is available in macOS 13+ so if we go native, we need to bump the lowest supported OS that was not an option, so I created my own version that for some reason you and depal had issues in the tests that I didn't so I switched to this Package, to make it available to macOS 12 too. As this feature, changes Sources JSON structure it will add tons of codes in `if available` conditions if we add it as an version dependent feature both in UI and logic.

For `Semaphore`, it blocks `asynchronous` tasks w/o freezing UI, like we have a `resolveSource` function that calls an `async` loading api function, so if we don't block the UI from updating w/o freezing it it will make duplicates of sources in the UI (not in logic), be my guest to suggest a workaround but personally I don't invent wheels.

Depal1:
> Can this be implemented without the Semaphore and NavigationStack-backport packages? It would be nice to not include too many external packages, as most of this can be done in pure Swift (as for NavigationStack, I was thinking it could be a feature restricted to the earliest MacOS version that supports it). @Depal1?

PlayCover runs fairly good in macOS 12, as a matter of fact it might be the best macOS version to run it (due to the obnoxious changes to Mac Catalyst since Ventura), so I am a bit adamant of cutting it out from the supported macOS version list if the reason behind the change is UI. Making `NavigationStack` related features macOS13+ only might be best instead of adding an additional package.

amirsaam:
> Making NavigationStack related features macOS13+ only might be best instead of adding an additional package.

@Depal1 you know that adds tons of `#if available` codes, hence I do volunteer to do it and add the fancy on top of it for macOS13+ for it too.
This approach to recreate the Stack in macOS12 was why you guys in first place asked for it.
I really tried to fix the issues but the behavior in different devices as you saw is outta control, I didn't have those issues on my device while you had.

amirsaam:
Closing this PR in order to recreate Source Folders as @Depal1 asked like macOS Finder.

Depal1 (approved):

Depal1 (changes requested):
The minimum width of the sidebar is not displaying the name 'IPA Library' completely since the dropdown arrow was added.

amirsaam:
> The minimum width of the sidebar is not displaying the name 'IPA Library' completely since the dropdown arrow was added.

@Depal1
I think your sidebar size has been changed manually before
We don't provide any value for the width of sidebar in codes so I cannot intervene in its behaviour.

amirsaam:
@Depal1 fixed.

Depal1 (approved):

TheMoonThatRises:
There seems to be an issue where switching from a source folder to the IPA Library will not display anything. Is this a reproducible issue @Depal1?

Depal1:
> There seems to be an issue where switching from a source folder to the IPA Library will not display anything. Is this a reproducible issue @Depal1?

I am not being able to reproduce it on my end. Switching from a source folder to IPA Library is displaying the IPAs from all sources.

PR #1715, opened 2024-10-14
Sources toggle

Catta1997:
image

Be able to activate and deactivate IPA sources

- [x] Toggles
- [x] Toggle text
- [x] List of enabled sources are saved and loaded when you open PlayCover

### DIFF
--- a/PlayCover/ViewModel/StoreVM.swift
+++ b/PlayCover/ViewModel/StoreVM.swift
@@ -20,6 +20,9 @@ class StoreVM: ObservableObject, @unchecked Sendable {
         resolveSources()
     }
 
+    @Published var enabledList: [String] = (UserDefaults.standard.stringArray(forKey: "enableSourceList") ??
+                                             StoreVM.shared.sourcesList.map { $0.source })
+
     @Published var sourcesList: [SourceData] {
         didSet {
             encode()
@@ -28,9 +31,10 @@ class StoreVM: ObservableObject, @unchecked Sendable {
     @Published var sourcesData: [SourceJSON] = [] {
         didSet {
             sourcesApps.removeAll()
-            for source in sourcesData {
-                appendSourceData(source)
-            }
+            let enabledSources: [SourceJSON] = sourcesData.filter { enabledList.contains($0.sourceURL) }
+              for source in enabledSources {
+                    appendSourceData(source)
+                }
         }
     }
     @Published var sourcesApps: [SourceAppsData] = []
@@ -39,12 +43,12 @@ class StoreVM: ObservableObject, @unchecked Sendable {
 
     //
     func enableSourceToggle(source: SourceData, value: Bool) {
-        if enableList.contains(source.source) && !value {
-            enableList.removeFirstObject(object: source.source)
+        if enabledList.contains(source.source) && !value {
+            enabledList.removeFirstObject(object: source.source)
         } else {
-            enableList.append(source.source)
+            enabledList.append(source.source)
         }
-        UserDefaults.standard.set(enableList, forKey: "enableSourceList")
+        UserDefaults.standard.set(enabledList, forKey: "enableSourceList")
     }
     //
     func addSource(_ source: SourceData) {
@@ -101,6 +105,23 @@ class StoreVM: ObservableObject, @unchecked Sendable {
         }
     }
 
+    /*
+    //
+    @MainActor func asyncresolveSources() async {
+        guard NetworkVM.isConnectedToNetwork() && !sourcesList.isEmpty else { return }
+        let sourcesCount = sourcesList.count
+        sourcesData.removeAll()
+        for index in sourcesList.indices where enabledList.contains(sourcesList[index].source) {
+            sourcesList[index].status = .checking
+            let (sourceJson, sourceState) = await getSourceData(sourceLink: sourcesList[index].source)
+            guard sourcesCount == sourcesList.count else { return }
+            sourcesList[index].status = sourceState
+            if sourceState == .valid, let sourceJson {
+                sourcesData.append(sourceJson)
+            }
+        }
+    }
+*/
     //
     func resolveSources() {
         resolveTask?.cancel()
@@ -111,7 +132,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
             let sourcesCount = sourcesList.count
             sourcesData.removeAll()
 
-            for index in sourcesList.indices where enableList.contains(sourcesList[index].source) {
+            for index in sourcesList.indices {
                 sourcesList[index].status = .checking
                 let (sourceJson, sourceState) = await getSourceData(sourceLink: sourcesList[index].source)
                 guard sourcesCount == sourcesList.count else { return }
@@ -178,7 +199,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
                 ? (url.absoluteString as NSString).lastPathComponent.replacingOccurrences(of: ".json", with: "")
                 : url.host ?? url.absoluteString
                 let oldTypeJson: [SourceAppsData] = try JSONDecoder().decode([SourceAppsData].self, from: unwrappedData)
-                decodedData = SourceJSON(name: sourceName, data: oldTypeJson)
+                decodedData = SourceJSON(name: sourceName, data: oldTypeJson, sourceURL: sourceLink)
                 return (decodedData, .valid)
             } catch {
                 debugPrint("Error decoding data from URL: \(url): \(error)")
@@ -200,6 +221,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
 struct SourceJSON: Codable, Equatable, Hashable {
     let name: String
     let data: [SourceAppsData]
+    let sourceURL: String
 }
 
 struct SourceAppsData: Codable, Equatable, Hashable {
--- a/PlayCover/ViewModel/StoreVM.swift
+++ b/PlayCover/ViewModel/StoreVM.swift
@@ -35,7 +35,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
 
     @Published var sourcesEnabledData: [SourceJSON] = [] {
         didSet {
-            sourcesApps.removeAll()
+            sourcesEnabeldApps.removeAll()
             for source in sourcesEnabledData {
                 appendSourceEnabledData(source)
             }
@@ -52,6 +52,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
     }
 
     @Published var sourcesApps: [SourceAppsData] = []
+    @Published var sourcesEnabeldApps: [SourceAppsData] = []
 
     private var resolveTask: Task<Void, Never>?
     //
@@ -223,8 +224,8 @@ class StoreVM: ObservableObject, @unchecked Sendable {
     }
     //
     private func appendSourceEnabledData(_ source: SourceJSON) {
-        for app in source.data where !sourcesApps.contains(app) {
-            sourcesApps.append(app)
+        for app in source.data where !sourcesEnabeldApps.contains(app) {
+            sourcesEnabeldApps.append(app)
         }
     }
 
--- a/PlayCover/ViewModel/StoreVM.swift
+++ b/PlayCover/ViewModel/StoreVM.swift
@@ -36,12 +36,12 @@ class StoreVM: ObservableObject, @unchecked Sendable {
     @Published var sourcesEnabledData: [SourceJSON] = [] {
         didSet {
             sourcesApps.removeAll()
-            for source in sourcesData {
-                appendSourceData(source)
+            for source in sourcesEnabledData {
+                appendSourceEnabledData(source)
             }
         }
     }
-    
+
     @Published var sourcesData: [SourceJSON] = [] {
         didSet {
             sourcesApps.removeAll()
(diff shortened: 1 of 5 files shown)


Catta1997:
Now when you toggle off a source it will greyed the text and the green checkmark will change (error symbol will remain the same as now).

https://github.com/user-attachments/assets/9ead9654-8297-4720-81f3-97727639409e

Also the enable source list will be populated with all sources when the decode() fail (first start or plist decode error)

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -10,7 +10,6 @@ import Foundation
 class StoreVM: ObservableObject, @unchecked Sendable {
     public static let shared = StoreVM()
     private let plistSource: URL
-
```
TheMoonThatRises: Small style nitpick, can you leave the newline here?

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -20,23 +19,43 @@ class StoreVM: ObservableObject, @unchecked Sendable {
...
-                appendSourceData(source)
-            }
+            updateSourcesApps()
         }
     }
     @Published var sourcesApps: [SourceAppsData] = []
 
     private var resolveTask: Task<Void, Never>?
 
+    //
```
TheMoonThatRises: No empty comments.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -20,23 +19,43 @@ class StoreVM: ObservableObject, @unchecked Sendable {
...
+    func updateSourcesApps() {
+        sourcesApps.removeAll()
+        let enabledSources: [SourceJSON] = sourcesData.filter { enabledList.contains($0.sourceURL) }
+          for source in enabledSources {
```
TheMoonThatRises: Inconsistent tab spacing here.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -201,3 +221,10 @@ struct SourceAppsData: Codable, Equatable, Hashable {
     let link: String
     let checksum: String?
 }
+
+extension Array where Element == String {
```
TheMoonThatRises: It would be better to use `.filter` instead of creating a function to do this.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -20,23 +19,43 @@ class StoreVM: ObservableObject, @unchecked Sendable {
...
+        let enabledSources: [SourceJSON] = sourcesData.filter { enabledList.contains($0.sourceURL) }
+          for source in enabledSources {
+                appendSourceData(source)
+            }
+    }
+
+    //
+    func enableSourceToggle(source: SourceData, value: Bool) {
+        if enabledList.contains(source.source) && !value {
+            enabledList.removeFirstObject(object: source.source)
```
TheMoonThatRises: Something like `enabledList.filter { $0 != source.source }` but might have to compare a specific value within SourceData.

PlayCover/Views/MainView.swift
```diff
@@ -59,7 +59,8 @@ struct MainView: View {
                             }
                         }
                         if showSourceFolders {
-                            ForEach(store.sourcesData, id: \.hashValue) { source in
+                            ForEach(store.sourcesData.filter {
+                                StoreVM.shared.enabledList.contains($0.sourceURL) }, id: \.hashValue) { source in
```
TheMoonThatRises: Inconsistent tab spacing here.

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -42,7 +42,8 @@ struct IPASourceSettings: View {
         Form {
             HStack {
                 List(storeVM.sourcesList, id: \.id, selection: $selected) { source in
-                    SourceView(source: source)
+                    SourceView(source: source,
+                               isEnabled: storeVM.enabledList.contains(where: { $0 == source.source }))
```
TheMoonThatRises: Now that I'm looking at this, it would be easier to have `StoreVM.enabledList` be a dictionary with the source name as the key and the value be if it is enabled or not. That would make the code cleaner.

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -102,11 +103,16 @@ struct IPASourceSettings: View {
 
 struct SourceView: View {
     var source: SourceData
+    @State var isEnabled: Bool
     @State var showingPopover = false
 
     var body: some View {
         HStack {
-            Text(source.source)
+            Toggle("", isOn: $isEnabled).onChange(of: isEnabled) { value in
```
TheMoonThatRises: `.onChange` should be on a newline after toggle, and preferably at the bottom of the view, as that is most common throughout the codebase. Also, the source text should be with the toggle instead of in its own `Text`.

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -102,11 +103,16 @@ struct IPASourceSettings: View {
...
+                StoreVM.shared.enableSourceToggle(source: source, value: value)
+            }
+            .help("state.enabled")
```
TheMoonThatRises, line 116: Is `state.enabled` already in the localisation file?

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -132,10 +138,17 @@ struct SourceView: View {
...
-                                showingPopover: $showingPopover)
+                if isEnabled {
+                    StatusBadgeView(imageName: "checkmark.circle.fill",
+                                    imageColor: .green,
+                                    popoverText: "preferences.popover.valid",
+                                    showingPopover: $showingPopover)
+                } else {
+                    StatusBadgeView(imageName: "checkmark.circle.badge.xmark.fill",
+                                    imageColor: .gray,
+                                    popoverText: "state.disabled",
```
TheMoonThatRises, line 150: Same here with localisation.

TheMoonThatRises (changes requested):
I haven't tested this code yet, but here are some quick feedback from a quick glance.

PlayCover/Views/Settings/IPASourceSettings.swift
Catta1997: Yes: https://github.com/PlayCover/PlayCover/blob/9a66b3a5ec0bec4f181529353045022acc9d6c57/PlayCover/ca.lproj/Localizable.strings#L243

PlayCover/Views/Settings/IPASourceSettings.swift
Catta1997: This is already in Localizable.strings; https://github.com/PlayCover/PlayCover/blob/9a66b3a5ec0bec4f181529353045022acc9d6c57/PlayCover/ca.lproj/Localizable.strings#L36

PlayCover/ViewModel/StoreVM.swift
Catta1997: I added it 'cause there are other "//" before func in that file

Catta1997:
Need to be tested after the change

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -11,23 +11,27 @@ struct SourceData: Identifiable, Hashable {
...
         try container.encode(source, forKey: .source)
+        try container.encode(isEnabled.description, forKey: .isEnabled)
     }
 }
 
 extension SourceData: Decodable {
     init(from decoder: Decoder) throws {
         let values = try decoder.container(keyedBy: SourceDataKeys.self)
         source = try values.decode(String.self, forKey: .source)
+        isEnabled = (try? values.decode(String.self, forKey: .isEnabled) == "true") ?? true
```
TheMoonThatRises: This should use `.decodeIfPresent` as it provides backwards compatibility, and the decoding type should be `Bool.self` instead of `String.self` to avoid the string comparison (requires change above to be made).

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -11,23 +11,27 @@ struct SourceData: Identifiable, Hashable {
...
     var status: SourceValidation = .valid
+    var isEnabled: Bool
 
     enum SourceDataKeys: String, CodingKey {
         case source
+        case isEnabled
     }
 }
 
 extension SourceData: Encodable {
```
TheMoonThatRises: Can you change the way `SourceData` encodes and decodes data to be the same as keymappings does it ([Model/Keymapping.swift](https://github.com/PlayCover/PlayCover/blob/develop/PlayCover/Model/Keymapping.swift))? This would make it easier to read without needing the use of `CodingKeys` while also preserving types.

TheMoonThatRises (changes requested):

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -132,12 +140,23 @@ struct SourceView: View {
...
+                    StatusBadgeView(imageName: "checkmark.circle.badge.xmark.fill",
+                                    imageColor: .gray,
+                                    popoverText: "state.disabled",
+                                    showingPopover: $showingPopover)
+                }
             }
         }
+        .onChange(of: isEnabled) { value in
+            StoreVM.shared.enableSourceToggle(source: source, value: value)
+            StoreVM.shared.updateSourcesApps()
```
TheMoonThatRises, line 157: When the user toggles the source while on the IPA Library tab, it causes PlayCover to stutter as it attempts to load and render all the apps. This should be fixed by simply putting this function inside a `Task`. You might need to store the task however to prevent weird behaviour when the user spams the toggle (maybe implement something similar to `resolveTask` in StoreVM. Or alternatively, you can disable the UI while the function is updating the available sources.

PlayCover/Views/Settings/IPASourceSettings.swift
```diff
@@ -8,26 +8,29 @@
...
     }
-}
 
-extension SourceData: Decodable {
     init(from decoder: Decoder) throws {
-        let values = try decoder.container(keyedBy: SourceDataKeys.self)
-        source = try values.decode(String.self, forKey: .source)
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.init(
+            id: try container.decode(UUID.self, forKey: .id),
```
TheMoonThatRises: `id` and `status` should also be `decodeIfPresent` for backwards compatibility. `status` shouldn't be read from the source file, but instead always be set to `.checking` by default.

TheMoonThatRises (changes requested):
This should be the last change before merging, PR works as intended.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -27,16 +27,32 @@ class StoreVM: ObservableObject, @unchecked Sendable {
...
+        sourcesApps.removeAll()
+        let enabledSources: [SourceJSON] = sourcesData.filter { sourceJSON in
+            return sourcesList.contains { sourceData in
+                sourceData.source == sourceJSON.sourceURL && sourceData.isEnabled
+            }
+        }
+        for source in enabledSources {
+            appendSourceData(source)
+        }
+}
```
TheMoonThatRises: Bracket needs tab spacing.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -191,6 +207,7 @@ class StoreVM: ObservableObject, @unchecked Sendable {
 struct SourceJSON: Codable, Equatable, Hashable {
     let name: String
     let data: [SourceAppsData]
+    let sourceURL: String
```
TheMoonThatRises: Instead of having `sourceURL` here, it could be replaced with `id` and that could be used to compare if the source is enabled or not instead of by url. This only really matters if people have duplicate source urls, but is a nice-to-have.

PlayCover/ViewModel/StoreVM.swift
```diff
@@ -27,16 +27,32 @@ class StoreVM: ObservableObject, @unchecked Sendable {
...
+    func enableSourceToggle(source: SourceData, value: Bool) {
+        if let index = sourcesList.firstIndex(of: source) {
+            sourcesList[index].isEnabled = value
+        }
+        updateSourcesApps()
+    }
+
+    func updateSourcesApps() {
+        sourcesApps.removeAll()
+        let enabledSources: [SourceJSON] = sourcesData.filter { sourceJSON in
```
TheMoonThatRises: It would be nice to have the `enabledSources` filter be a function as this exact code is used multiple times.

PlayCover/Views/Settings/IPASourceSettings.swift
Catta1997:
```
init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        self.init(
            id: try container.decodeIfPresent(UUID.self, forKey: .id) ?? UUID(),
            source: try container.decode(String.self, forKey: .source),
            status: .checking,
            isEnabled: try container.decodeIfPresent(Bool.self, forKey: .isEnabled) ?? true
        )
    }
```

Like this?

TheMoonThatRises: Yes.

TheMoonThatRises (approved):

PR #1752, opened 2024-11-28
Specific and exhaustive decompression errors

Catta1997:
Specific and exhaustive error when a file reach a google drive quota limit, before was a long error that was difficult to understand 

Screenshot 2024-11-28 alle 09 09 32

I'm working on a auto-retry when this happens

### DIFF
--- a/PlayCover/AppInstaller/Installer.swift
+++ b/PlayCover/AppInstaller/Installer.swift
@@ -118,10 +118,12 @@ class Installer {
                 InstallVM.shared.next(.finish, 0.95, 1.0)
                 returnCompletion(finalURL)
             } catch {
-                Log.shared.error(error)
-
                 ipa.releaseTempDir()
-
+                if error.localizedDescription.contains("End-of-central-directory signature not found") {
+                    Log.shared.error(NSLocalizedString("alert.quota.limit", comment: ""))
+                } else {
+                    Log.shared.error(error)
+                }
                 InstallVM.shared.next(.failed, 0.95, 1.0)
                 returnCompletion(nil)
             }
--- a/PlayCover/AppInstaller/Installer.swift
+++ b/PlayCover/AppInstaller/Installer.swift
@@ -35,6 +35,19 @@ class Installer {
         return response == .alertFirstButtonReturn
     }
 
+    static private func returnError(error: Error, googleDrive: Bool) {
+        switch error.localizedDescription {
+        case let str where str.contains("End-of-central-directory") && googleDrive:
+            Log.shared.error(NSLocalizedString("alert.quota.limit", comment: ""))
+        case let str where str.contains("End-of-central-directory") && !googleDrive:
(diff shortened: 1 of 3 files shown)


Catta1997:
I think its all

PlayCover/AppInstaller/Installer.swift
```diff
@@ -35,8 +35,22 @@ class Installer {
         return response == .alertFirstButtonReturn
     }
 
+    static private func returnError(error: Error, googleDrive: Bool) {
+        switch error.localizedDescription {
+        case let str where str.contains("End-of-central-directory") && googleDrive:
```
TheMoonThatRises: Instead of checking against a boolean `googleDrive`, quota limit can be found by checking if the string also contains `.html` if I recall correctly. See if that is possible, as other websites with quota limits (that are not Google Drive) may also return an HTML.

PlayCover/en.lproj/Localizable.strings
```diff
@@ -289,4 +289,6 @@
 "macos.version" = "This app can be installed from Mac App Store";
 "alert.install.anyway" = "Install Anyway";
 "alert.open.appstore" = "Open Mac App Store";
-
+"alert.quota.limit" = "Google Drive quota limit reached. It may take up 24 hours to download this file again";
```
TheMoonThatRises: If the comment above is possible, this could be changed to just quota limit reached without specifying Google Drive.

PlayCover/AppInstaller/Installer.swift
```diff
@@ -118,7 +131,7 @@ class Installer {
                 InstallVM.shared.next(.finish, 0.95, 1.0)
                 returnCompletion(finalURL)
             } catch {
-                Log.shared.error(error)
```
TheMoonThatRises, line 121: It would be easier to read if the `returnError` function returned the localised string and this is kept to `Log.shared.error`.

Catta1997: So `returnError(error) -> String` and `Log.shared.error(returned_value)` ?

TheMoonThatRises: Yep.

PlayCover/AppInstaller/Installer.swift
```diff
@@ -35,6 +35,19 @@ class Installer {
         return response == .alertFirstButtonReturn
     }
 
+    static private func returnErrorString(error: Error) -> String {
+        switch error.localizedDescription {
+        case let str where str.contains("(disk full?)"):
+            return NSLocalizedString("alert.notSpace", comment: "")
```
TheMoonThatRises: The switch statement doesn't need a `return` in it as by default it will return the value [SE-0380](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md).

TheMoonThatRises (approved):

PR #1762, opened 2024-12-07
MacOS detect

Catta1997:
Changed how PlayCover detect if an app has macOS version, in this way is no more necessary to add manually bundleID

### DIFF
--- a/PlayCover/Utils/IPA.swift
+++ b/PlayCover/Utils/IPA.swift
@@ -119,9 +119,8 @@ public class IPA {
             appID = Int(stringArray.last ?? "0") ?? 0
         }
         let supportMacOS: Bool = await checkMacOSCompatibility(appID: appID)
-        let noMacAlert = UserDefaults.standard.bool(forKey: "\(bundleID).noMacAlert")
         let showAlert = InstallPreferences.shared.showAppStorePopup
-        if showAlert && supportMacOS && !noMacAlert {
+        if showAlert && supportMacOS {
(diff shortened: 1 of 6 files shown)


PlayCover/Utils/IPA.swift
```diff
@@ -87,21 +118,29 @@ public class IPA {
             let stringArray = appLookup.components(separatedBy: CharacterSet.decimalDigits.inverted)
             appID = Int(stringArray.last ?? "0") ?? 0
         }
+        let supportMacOS: Bool = await checkMacOSCompatibility(appID: appID)
         let noMacAlert = UserDefaults.standard.bool(forKey: "\(bundleID).noMacAlert")
-        if PlayApp.MACOS_APPS.contains(bundleID), !noMacAlert {
+        let showAlert = InstallPreferences.shared.showAppStorePopup
+        if showAlert && supportMacOS && !noMacAlert {
             let alert = NSAlert()
-            alert.messageText = NSLocalizedString("alert.error", comment: "")
```
TheMoonThatRises, line 93: The alert message text should be something along the lines of `MacOS version available`.

TheMoonThatRises (changes requested):

PlayCover/Utils/IPA.swift
Catta1997: image

Like this?

TheMoonThatRises (approved):

PR #1768, opened 2024-12-11
Add support for iPhone 16 Pro Max

hyufe:
also see: https://github.com/PlayCover/PlayTools/pull/176


TheMoonThatRises:
Are there any measurable performance/compatibility improvements adding the iPhone 16 Pro Max model?

hyufe:
@TheMoonThatRises Sorry, I can't perform testing as I'm not very familiar with the playcover codebase. I added iPhone 16 Pro Max support because I play a game that only supports 120fps on the iPhone 16 Pro Max model. If you have plans to support iPhone 16 Pro Max in the future, please close this PR.

TheMoonThatRises (approved):

PR #1807, opened 2025-01-25
Condense supported Mac variants in README

SoCuul:


amirsaam:
I would prefer a "M1 and later devices" sentence instead of listing all M-series chip names.

XuYicong:
> I would prefer a "M1 and later devices" sentence instead of listing all M-series chip names.

Yeah it's better this way. The reason why it's now listing all models is a historic thing, when PlayCover just started there were only M1, M1 pro and M1 max, so they just directly wrote these model names. It seems silly if we continue listing all model names.

SoCuul:
Hope this phrasing works better.

TheMoonThatRises (approved):

PR #1814, opened 2025-01-28
Fix empty async image url string

TheMoonThatRises:
Provides a layer between `CachedAsyncImage` to check for `nil` and empty input url strings.

### DIFF
--- a/PlayCover/Views/CachedAsyncImageWrapper.swift
+++ b/PlayCover/Views/CachedAsyncImageWrapper.swift
@@ -1,0 +1,28 @@
+//
+//  CachedAsyncImageWrapper.swift
+//  PlayCover
+//
+//  Created by TheMoonThatRises on 1/28/25.
+//
+
+import CachedAsyncImage
+import SwiftUI
+
+struct CachedAsyncImageWrapper: View {
+
+    let url: String?
+    let placeholder: ((String) -> any View)?
+    let image: (CPImage) -> any View
+    let error: ((String, @escaping () -> Void) -> any View)?
+
+    var body: some View {
+        if let url = url, !url.isEmpty {
+            CachedAsyncImage(url: url, placeholder: placeholder, image: image, error: error)
+        } else {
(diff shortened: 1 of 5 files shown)


PlayCover/Views/CachedAsyncImageWrapper.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Created by TheMoonThatRises on 1/28/25.
+//
+
+import CachedAsyncImage
+import SwiftUI
+
+struct CachedAsyncImageWrapper: View {
+
+    let url: String?
+    let placeholder: ((String) -> any View)?
```
amirsaam, line 14: isn't the whole reason of creating this wrapper to make sure there is an image url to grab and show?
making it optional and also another view is not good approach from my point of view.
maybe adding a placeholder image to assets and load that as an image a better way to handle it instead.

TheMoonThatRises: The reason for the wrapper is to handle empty or nil url inputs, as the `CachedAsyncImage` library cannot handle it and will return an error into the console when attempting to resolve a nil or empty string.

amirsaam: Show a ProgressView instead of creating an empty input.

TheMoonThatRises: It doesn't create an empty view, it uses the `placeholder` provided to the original `CachedAsyncImage` if it is unable to resolve the url. This wrapper should provide the same behavior as before, but this wrapper simply makes it so the package doesn't log an error and provide undefined behavior if the url is nil or empty.

amirsaam: if it doesn't change behaviour, then lgtm

amirsaam (approved):

JoseMoreville (approved):

PR #1862, opened 2025-04-28
Check New Version on app start

Catta1997:
This PR add a check before the launch with an alert if there is a new version available in one of the available sources

### DIFF
--- a/PlayCover/Utils/VersionCheck.swift
+++ b/PlayCover/Utils/VersionCheck.swift
@@ -45,9 +45,9 @@ class VersionCheck {
         StoreVM.shared.resolveSources()
         let storeApp = StoreVM.shared.sourcesApps
         if let app = storeApp.first(where: {$0.bundleID == myApp.info.bundleIdentifier}) {
-            switch myApp.info.bundleVersion.compare(app.bundleID, options: .numeric) {
-            case .orderedDescending:
                 return await checkUpdateAlert(app: app)
+            switch myApp.info.bundleVersion.compare(app.version, options: .numeric) {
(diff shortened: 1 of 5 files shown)


PlayCover/Utils/VersionCheck.swift
```diff
@@ -0,0 +1,56 @@
...
+            return true
+        }
+    }
+
+    func checkNewVersion(myApp: PlayApp) async -> Bool {
+        StoreVM.shared.resolveSources()
+        let storeApp = StoreVM.shared.sourcesApps
+        if let app = storeApp.first(where: {$0.bundleID == myApp.info.bundleIdentifier}) {
+                return await checkUpdateAlert(app: app)
+            switch myApp.info.bundleVersion.compare(app.version, options: .numeric) {
```
TheMoonThatRises: Unclear what is happening here. Is the `switch` statement supposed to be here?

PlayCover/Utils/VersionCheck.swift
```diff
@@ -0,0 +1,54 @@
...
+                    let redirectHandler = RedirectHandler(url: url) // checking page redirect
+                    DownloadApp(url: redirectHandler.getFinal(), app: app, warning: nil).start()
+                }
+            }
+            return true
+        default:
+            return true
+        }
+    }
+
```
TheMoonThatRises, line 43: There is an edge-case where if the user loads PlayCover and immediately launches an app, the sources have not fully resolved and thus the version check does not work.

TheMoonThatRises (approved):

PR #1917, opened 2025-07-07
Feature: optional feature to hide titlebar

JoseMoreville:
This pull request introduces a new feature to hide the title bar in the application settings. The changes span multiple files to implement the feature, update the user interface, and add localization support.

### Feature Implementation:

* [`PlayCover/Model/AppSettings.swift`](diffhunk://#diff-3f57cf6f40ef6e282ae2799f270543767060ac9405adaaf8a2ecc1f89d79d269R44): Added a new property `hideTitleBar` to the `AppSettingsData` struct and updated the decoding logic to handle this property with a default value of `false`. [[1]](diffhunk://#diff-3f57cf6f40ef6e282ae2799f270543767060ac9405adaaf8a2ecc1f89d79d269R44) [[2]](diffhunk://#diff-3f57cf6f40ef6e282ae2799f270543767060ac9405adaaf8a2ecc1f89d79d269R75)

### User Interface Updates:

* `PlayCover/Views/AppSettingsView.swift`: 
  - Introduced a new `@AppStorage` binding for `hideTitleBar` in the `GraphicsView` struct.
  - Added a toggle UI element for `hideTitleBar` in the settings view, allowing users to enable or disable the feature.

### Localization:

* [`PlayCover/en.lproj/Localizable.strings`](diffhunk://#diff-c81219d3648ed083c147833f2b91fee16fa3e375249183f98a8eb8e2080970e4R171): Added a localized string for the new toggle option, "Hide titlebar".

related to: https://github.com/PlayCover/PlayTools/pull/191

### DIFF
--- a/PlayCover/Views/AppSettingsView.swift
+++ b/PlayCover/Views/AppSettingsView.swift
@@ -188,13 +188,12 @@ struct KeymappingView: View {
 
 struct GraphicsView: View {
     @Binding var settings: AppSettings
-
     @State var customWidth = 1920
     @State var customHeight = 1080
-
     @State var showResolutionWarning = false
     @AppStorage("settings.settings.inverseScreenValues") private var inverseScreenValues = false
     @AppStorage("settings.settings.disableTimeout") private var disableTimeout = false
+    @AppStorage("settings.toggle.hideTitleBar") private var hideTitleBar = false
     static var number: NumberFormatter {
         let formatter = NumberFormatter()
         formatter.numberStyle = .none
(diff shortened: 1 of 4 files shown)


JoseMoreville:
@TheMoonThatRises Will squash and merge, not needed to be done in pr

PlayCover/Views/AppSettingsView.swift
```diff
@@ -272,11 +271,8 @@ struct GraphicsView: View {
                                 })
                                 .frame(width: 125)
                         }
-                        onIncrement: {
```
TheMoonThatRises, line 275: Is removing these new lines to ensure that the file is within the file length limit? If so, I think we should either add a SwiftLint exception for the file or find a way to split this into multiple files.

JoseMoreville: > I think the best thing is to increase the length limit and modify the rules, 250 is too low

And yes is to avoid the limit.

TheMoonThatRises (approved):

PR #1966, opened 2025-09-13
Feature: optional feature to floating window

Undefined-User:
Added setting options to keep the application window in front of other windows https://github.com/PlayCover/PlayTools/pull/198

### DIFF
--- a/PlayCover/Model/AppSettings.swift
+++ b/PlayCover/Model/AppSettings.swift
@@ -42,6 +42,7 @@ struct AppSettingsData: Codable {
     var noKMOnInput = true
     var enableScrollWheel = true
     var hideTitleBar = false
+    var floatingWindow = false
     var checkMicPermissionSync = false
 
     init() {}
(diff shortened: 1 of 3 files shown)


Undefined-User:
PR is ready for review, but I can't add reviewer myself. @TheMoonThatRises @JoseMoreville

Catta1997:
Can you check if this fix also the problem with macOS tahoe and controller? When app is not focused controller no longer work

Undefined-User:
> Can you check if this fix also the problem with macOS tahoe and controller? When app is not focused controller no longer work

Sorry, I haven't updated to the latest version of macOS yet. I don't know much about the background information of the controller issue, but if you can find a qualified tester, I can provide a compiled binary using this patch.

JoseMoreville:
If someone can test this, it would be appreciated.
> Can you check if this fix also the problem with macOS tahoe and controller? When app is not focused controller no longer work

TheMoonThatRises (approved):

PR #1967, opened 2025-09-13
Fix: Unable to install unsigned IPA

### LINKED ISSUE #1605: [Feature]: Allow opening of unsigned ipa

I was trying to open certain apps, however I get the popup of : code object is not signed at all, because they were not signed. There was one issue related to this, someone needing help with "Arceus X".

I would like to have an option to be able to run unsigned ipa files with a warning of it being unsigned and possibly dangerous or risky.

### Anything else?

If you do not want to add this feature due to safety, I would like to know how I could fork and modify the project to allow unsigned ipas to be run.

### Issue Language

Undefined-User:
Related issue: https://github.com/PlayCover/PlayCover/issues/1605

Note: Unsigned IPA will works fine, but some IPA files (In my case it had been signed by a revoked enterprise certificate) framework binary may return `internal error in Code Signing subsystem` when executing `/usr/bin/codesign -fs- framework_binary` in `PlayCover/Utils/Shell.swift`. If we ignore error above, app will crash with runtime error `code signature invalid in <XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX>`. I haven't found a solution for this case yet.

### DIFF
--- a/PlayCover/Utils/PlayTools.swift
+++ b/PlayCover/Utils/PlayTools.swift
@@ -230,6 +230,9 @@ class PlayTools {
             if error.localizedDescription.contains("Document is empty") {
                 // Empty entitlements
                 return ""
+            } else if error.localizedDescription.contains("code object is not signed at all") {
+                // IPA not signed
+                return ""
             } else {
                 throw error
             }


Undefined-User:
PR is ready for review, but I can't add reviewer myself. @TheMoonThatRises @JoseMoreville

TheMoonThatRises (approved):

PR #2001, opened 2025-10-15
Fix build step for local PlayTools repository

Undefined-User:
I modified `Cartfile` file to `git "file:///Users/PATH_IN_LOCAL/PlayTools" "master"`, hoping to use the git commit in the local PlayTools repository when compiling the PlayCover source code.

However, the environment variables passed by Xcode 15.2 caused the `carthage update --cache-builds --use-xcframeworks` command to fail, so I executed the carthage command in a new terminal session without passing the environment variables.

### DIFF
--- a/PlayCover.xcodeproj/project.pbxproj
+++ b/PlayCover.xcodeproj/project.pbxproj
@@ -629,7 +629,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


Undefined-User:
PR is ready for review @TheMoonThatRises

irl231:
I also encountered this when using forked Playtools repo, I've tried to change the permission of Cartfile to allow write access and give full disk access to XCode still get permission denied. That's why I didn't modify `Cartfile`.

JoseMoreville:
This should be only for forks, not for the prod or dev one under playcover org.

TheMoonThatRises:
I think a better solution would be to have the command run in an isolated environment without the use of `sudo`. The use of `sudo` is unappealing regardless of if it is running an escalated privilege command.

Undefined-User:
> I think a better solution would be to have the command run in an isolated environment without the use of `sudo`. The use of `sudo` is unappealing regardless of if it is running an escalated privilege command.

Ok, I changed the command to `env -i` and appended the necessary environment variables.

Undefined-User:
> This should be only for forks, not for the prod or dev one under playcover org.

Yes, this is a change that applies to forked versions, but will not affect existing developer of the PlayCover project. I hope this change will make it easier for people who want to contribute code to the PlayCover project to test their changes locally.

TheMoonThatRises (approved):

PR #2011, opened 2025-10-25
Add better rotation management options

luka2272:
Instead of having to keep pressing the display rotation button I wanted to add a way to both change the default orientation an application launches with and instead of a button to cycle through orientations, have a list from which you can pick a specific orientation. 

https://github.com/PlayCover/PlayTools/pull/203


PlayCover.xcodeproj/project.pbxproj
JoseMoreville, line 1: Remove this from your PR please.

PlayCover/Utils/UpdateScheme.swift
JoseMoreville, line 1: Is this necessary?

Cartfile
JoseMoreville, line 1: Please do not point to your branch of pT

JoseMoreville (dismissed):

PlayCover/Utils/UpdateScheme.swift
luka2272: Yes. I found that the program gets stuck in a loop trying to upgrade version 3.1 to 3.1, failing and crashing if I don't make this change.

TheMoonThatRises: That is outside the scope of this pull request. Please submit a new pull request that is dedicated for this fix.

TheMoonThatRises:
Can you remove all the localizations other than English? Those should only be done through Weblate whenever we get that back open.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -193,8 +193,10 @@ struct GraphicsView: View {
     @State var showResolutionWarning = false
     @AppStorage("settings.settings.inverseScreenValues") private var inverseScreenValues = false
     @AppStorage("settings.settings.disableTimeout") private var disableTimeout = false
+    @AppStorage("settings.settings.keepDisplayRotation") private var keepDisplayRotation = true
```
TheMoonThatRises: This seems to be an extraneous variable.

PlayCover/Views/AppSettingsView.swift
```diff
@@ -351,6 +353,19 @@ struct GraphicsView: View {
                         }
                         Spacer()
                     }
+                    HStack {
+                        Text("settings.settings.displayRotation")
+                        Spacer()
+                        Picker("", selection: $settings.settings.displayRotation) {
```
TheMoonThatRises, line 358: The text in the picker should be localized.

TheMoonThatRises (changes requested):

PlayCover/en.lproj/Localizable.strings
```diff
@@ -175,6 +175,12 @@
 "settings.picker.windowFixMethod.0" = "Normal";
 "settings.picker.windowFixMethod.1" = "Alternate";
 "settings.toggle.disableDisplaySleep" = "Disable display sleep";
+"settings.settings.displayRotation" = "Device Orientation";
```
TheMoonThatRises, line 178: Can you make the localization keys have the prefix `settings.settings.displayRotation`? For example, `settings.settings.default` would be `settings.settings.displayRotation.default`. This would make the key more self-descriptive.

PlayCover/Views/MainView.swift
```diff
@@ -76,7 +76,7 @@ struct MainView: View {
                             }
                         }
                     }
-                    .frame(minWidth: 150)
+                    .frame(minWidth: 140)
```
TheMoonThatRises: Is there a particular reason for changing the minimum frame width to 140?

TheMoonThatRises (changes requested):

PlayCover/Views/MainView.swift
luka2272: For some reason if it's 150, the sidebar items get cut off and you have to manually resize the sidebar to fix it. Setting it to 140 fixes the issue.

Depal1 (approved):

TheMoonThatRises (approved):

JoseMoreville (approved):

PR #2026, opened 2025-11-08
Fix crash/race condition

enginelesscc:
Fixes a crash due to unsynchronized access

Im not a swift dev - I just needed to fix this crash.
Though I may as well share :)

### DIFF
--- a/PlayCover/AppInstaller/Installer.swift
+++ b/PlayCover/AppInstaller/Installer.swift
@@ -172,6 +172,7 @@ class Installer {
         }
 
         var resolved: [URL] = []
+        let serialQueue = DispatchQueue(label: "baseAppUrlResolver")
 
         baseApp.url.enumerateContents { url, attributes in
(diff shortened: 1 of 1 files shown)


PlayCover/AppInstaller/Installer.swift
```diff
@@ -172,6 +172,7 @@ class Installer {
         }
 
         var resolved: [URL] = []
+        let serialQueue = DispatchQueue(label: "baseAppUrlResolver")
```
TheMoonThatRises, line 175: Is it possible to use Swift 5 Async/Await concurrency instead of the old `DispatchQueue`? We have somewhat recently moved the code base away from using `DispatchQueue`.

enginelesscc: I personally would have just preferred a simple mutex/lock. Concurrency alone does not guarantee data race safety to me. I was quite surprised Swift does not natively offer a simple mutex type...

TheMoonThatRises (approved):

Issue #2051, opened 2025-12-31
[Bug]: PlayCover_nightly_1346 Cannot install ipa


eachann1024:
### Describe the bug

<img width="887" height="449" alt="Image" src="https://github.com/user-attachments/assets/742034b6-5e3b-4670-87ae-4b19d930e057" />

提示
未能将 "jkchess.app”移到
“Applications”，因为前者不存在， 或包含后者的文件夹不存在。

尝试过 chatgpt 多种方案修改都不行

设备信息:
Apple M4 Pro  mac: 26.2 (25C56)
尝试安装多个每夜版本都不行

draw attention to sth.

Failed to move "jkchess.app" to the

"Applications" because the former does not exist, or the folder containing the latter does not exist.

Tried chatgpt with several options but no luck!

Device Info.

Apple M4 Pro mac: 26.2 (25C56)

Tried installing multiple nightly versions but no luck

### Steps to reproduce

1. 引入源
2. 下载金铲铲之战
3. 安装助手后最后一步提示

1. Introducing sources
2. Download Battle of the Golden Shovels
3. Install the assistant after the last step prompt

### Expected behaviour

1. 引入源
2. 下载金铲铲之战
3. 安装助手后最后一步提示

1. Introducing sources
2. Download Battle of the Golden Shovels
3. Install the assistant after the last step prompt

### Crash log

```shell

```

### What version of PlayCover are you using?

Nightly/beta (please specify build number)

### What version of macOS are you using?

Other (please specify)

### Issue Language

- [x] Yes my issue is written in English

TheMoonThatRises:
Try running `mkdir -r ~/Library/Containers/io.playcover.PlayCover/Applications` and install again.

eachann1024:
> Try running `mkdir -r ~/Library/Containers/io.playcover.PlayCover/Applications` and install again.

No, it's still the same problem.

Issue #2085, opened 2026-02-20
[App Support]: V26.4 Beta Tahoe


JRodriguez-Luna:
### App Store link

 https://apps.apple.com/us/app/honkai-star-rail/id1599719154

### Current issues

- [x] App fails to install
- [ ] App fails to open
- [ ] App crashes randomly during usage
- [ ] App experiences graphical issues or broken UI
- [ ] App fails to connect to servers
- [ ] Other

### More details

When attempting to install the IPA, signing completes successfully but PlayCover fails to move hkrpg.app to /Applications with the error: "hkrpg couldn't be moved to Applications because either the former doesn't exist, or the folder containing the latter doesn't exist." Running macOS 26.4 beta (Build 25E5207k). The /Applications folder exists and has correct permissions (drwxrwxr-x). SIP is enabled.

### Crash log

```shell
Version 26.4 (Build 25E5207k)
Installing PlayTools
Copying PlayTools to Frameworks

Executable=/private/var/folders/qw/ky5gkqc906155f4qnll3m5w40000gn/T/TemporaryItems/NSIRD_PlayCover_GLecs5/Payload/hkrpg.app/hkrpg
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd"><plist version="1.0"><dict><key>application-identifier</key><string>5JG79VMBAP.com.HoYoverse.hkrpgoversea</string><key>aps-environment</key><string>production</string><key>com.apple.developer.applesignin</key><array><string>Default</string></array><key>com.apple.developer.associated-domains</key><array><string>applinks:oig.hoyoverse.com/hkrpgoversea</string><string>applinks:jtht.adj.st</string><string>applinks:oig.hoyoverse.com</string></array><key>com.apple.developer.game-center</key><true/><key>com.apple.developer.kernel.extended-virtual-addressing</key><true/><key>com.apple.developer.kernel.increased-memory-limit</key><true/><key>com.apple.developer.team-identifier</key><string>5JG79VMBAP</string><key>keychain-access-groups</key><array><string>5JG79VMBAP.com.HoYoverse.hkrpgoversea</string><string>5JG79VMBAP.com.miHoYo.devicefp</string></array></dict></plist>

[...]

Issue #2004, opened 2025-10-18
[Bug]: Update function in dev branch version creates endless loop and crashes the program


luka2272:
### Describe the bug

While trying to run the latest nightly version of PlayCover, I found it sat in the dock with a not responding status. Running it through the terminal shows it is stuck in a checking for updates loop, while trying to update PlayCover from version 3 to 3.1. I had the latest 3.1 stable version installed so this update thing seemed strange. Looking through the code I found that the responsible function for the 3 to 3.1 update is in a switch that checks for version "3", regardless of the point release, thus trying to update version 3.1 to... 3.1 and fails. Modifying the string to check for strictly "3.0" instead of "3" fixes the issue and the app starts right up. Didn't do a PR since I thought this was a minor thing, even if without this change the app is unusable:) Hope it gets fixed!

### Steps to reproduce

Using macOS Tahoe 26.1 Beta 2, Xcode 26.0.1 and the latest development code at the time of writing. Just git cloning and compiling the existing code should show the issue. I don't think the os version is really relevant in this case but I'd be curious to know if it is.

### Expected behaviour

The app should check for a version string of 3.0 when trying to update version 3 to 3.1 as to not try updating version 3.1 to 3.1.

### Crash log

```shell

```

### What version of PlayCover are you using?

Nightly/beta (please specify build number)

### What version of macOS are you using?

macOS beta (please state the specific version)

### Issue Language

- [x] Yes my issue is written in English

naley2300:
Please show me how to fix the same problem above. Where to modify the string to check version stricly to 3.0

luka2272:
In UpdateScheme.swift, in the checkForUpdate() function, in the switch case instruction, the case for "3" is the one to modify.

Issue #1377, opened 2024-02-05
[Bug]: Failed to move “hkrpg” to “Applications”，because the former doesn't exist，or the directory containing the latter doesn't exist.


ellplk:
### Describe the bug

安装软件时遇到以上问题

### Steps to reproduce

1.install playcover 3.0.2
2.download hkrpg.ipa file
3.import file

### Expected behaviour

无

### What version of PlayCover are you using?

3.0.0 beta 2

### What version of macOS are you using?

Sonoma (macOS 14)

### Issue Language

- [X] Yes my issue is written in English

scpopcorn13:
Is this bug fixed? I am in the same situation.
<img width="372" alt="截屏2024-12-26 22 52 01" src="https://github.com/user-attachments/assets/071b6a21-3711-43e0-b22b-1edb7e6ee966" />

luke358:
2.0.5 is ok

cjackey513:
I also encountered the same problem

imcdd:
运行终端执行以下命令（execute this command on terminal）
mkdir /Users/$(whoami)/Library/Containers/io.playcover.PlayCover/Applications

KrisTumYYYes:
> 运行终端执行以下命令（execute this command on terminal） mkdir /Users/$(whoami)/Library/Containers/io.playcover.PlayCover/Applications
Does "$(whoami)" need to be replaced by my name ?

6or7mangos:
> > 运行终端执行以下命令（execute this command on terminal） mkdir /Users/$(whoami)/Library/Containers/io.playcover.PlayCover/Applications
> > Does "$(whoami)" need to be replaced by my name ?

whoami acts as a universal fill in for the mac usernames,so no

PR #2092, opened 2026-02-26
fix: folder and update fix

Catta1997:
- Fix 'Applications' folder not created on Nightly
- Fix Infinite Recursion reported on a Discord support thread

### DIFF
--- a/PlayCover/Utils/UpdateScheme.swift
+++ b/PlayCover/Utils/UpdateScheme.swift
@@ -10,25 +10,58 @@ class UpdateScheme {
     public static var currentVersion: String {
         (try? String(contentsOf: UpdateScheme.versionsFile)) ?? "3.1"
     }
+    struct Version: Hashable {
+        let components: [Int]
+        init(_ string: String) {
+            var comps = string.split(separator: ".").compactMap { Int($0) }
+            while comps.last == 0 && comps.count > 1 {
+                comps.removeLast()
+            }
+            self.components = comps
+        }
+    }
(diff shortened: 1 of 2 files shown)


Catta1997:
Probably will fix #2051 , #1377, #2085 and others

PlayCover/Utils/UpdateScheme.swift
```diff
@@ -29,7 +29,6 @@ class UpdateScheme {
             Log.shared.error(error)
         }
 
-        checkForUpdate()
```
TheMoonThatRises, line 32: This recursive call is necessary for users who are updating through multiple versions (e.g. from v2 to v3.1). There was an issue that noted a fix for the specific recursive crash (#2004). However, it would be nice to add some sort of check to ensure that this recursion only happens a maximum number of times before erroring out and continuing.

PlayCover/ViewModel/AppsVM.swift
```diff
@@ -12,10 +12,18 @@ class AppsVM: ObservableObject {
     static let shared = AppsVM()
 
     private init() {
+        try? Self.ensureBaseDirectoriesExist()
```
TheMoonThatRises: Most of the code in the repository utilizes the class name over `Self` (i.e. replace `Self` with `AppsVM`).

TheMoonThatRises (changes requested):

TheMoonThatRises (approved):

PR #2096, opened 2026-03-06
fix: sandbox leaks for certain apps

ohaiibuzzle:
Removed:
- Sandbox exception for CookieRunKingdom (wat)
- Sandbox profiles of BlueArchive because I am not even sure if they still runs. Please comment if they still do.
- Sandbox leaks certain basic paths.

PlayCover should now pass basic Jailbreak checks by default. Advanced checks WILL still fail, however. Exercise caution.


TheMoonThatRises (changes requested):
The deleted YAML files need to be removed from the "Copy Bundle Resources" build phase. Along with that, the references should be removed via Xcode (or manually removed) as they still exist within the `project.pxbproj` file.

ohaiibuzzle:
795ea04 should do it.

TheMoonThatRises (approved):
Funnily enough, removing the rules for Blue Archive now makes it able to launch, as previously it was unable to access the updated keymapping system.

PR #2101, opened 2026-03-17
feat: Migrated ToastView visual style to the new design system

Stv-X:
Pleased to notice that this project has begun building against the macOS 26 SDK, allowing most visual styles to automatically benefit from the new design system. This Pull Request provides supplementary updates specifically for the custom `ToastView`.

1. **Adoption of Liquid Glass:** For macOS 26 and later, `ToastView` now utilizes [`Glass`](https://developer.apple.com/documentation/swiftui/glass) to better align with macOS's [Liquid Glass](https://developer.apple.com/documentation/swiftui/applying-liquid-glass-to-custom-views) design system. On earlier versions, the original [`Material`](https://developer.apple.com/documentation/SwiftUI/Material) style is preserved.

compare

2. **Safe Area Bar:** For macOS 26 and later, the `overlay` modifier used for displaying the `ToastView` in `MainView` has been replaced with the new [`safeAreaBar`](https://developer.apple.com/documentation/swiftui/view/safeareabar(edge:alignment:spacing:content:)), which provides an elegant progressive blur effect at the edges of the scroll view.

SafeAreaBar

3. **ContainerRelativeShape:** The shape of the Toast is now using [`ContainerRelativeShape`](https://developer.apple.com/documentation/SwiftUI/ContainerRelativeShape) instead of hardcoded, fixed corner radius values. This ensures [concentricity](https://developer.apple.com/videos/play/wwdc2025/356?time=183) between the view and the window corners. Given that this API has been available since macOS 11, this improvement benefits PlayCover across all supported OS versions.

concentric

### DIFF
--- a/PlayCover/Utils/Extensions/ViewExtensions.swift
+++ b/PlayCover/Utils/Extensions/ViewExtensions.swift
@@ -1,5 +1,5 @@
 //
-//  URLExtensions.swift
+//  ViewExtensions.swift
 //  PlayCover
 //
 
--- a/PlayCover/Utils/Extensions/ViewExtensions.swift
+++ b/PlayCover/Utils/Extensions/ViewExtensions.swift
@@ -16,11 +16,20 @@ extension View {
     }
 
     func toastBackground() -> some View {
+        let view = self
+            .padding()
+            .frame(maxWidth: .infinity)
+
+        // provide default padding for older systems, but for those with liquid glass available, add proper padding to
+        // left, right, and bottom to ensure that the distance from the edges of the app are consistent and equal
         #if compiler(>=6.2)
         if #available(macOS 26.0, *) {
-            return self.glassEffect(.regular, in: .containerRelative)
+            return view.glassEffect(.regular, in: .containerRelative)
+                .padding(ToastView.toastGlassPadding)
+                .padding(.top)
         }
         #endif
-        return self.background(.regularMaterial, in: .containerRelative)
+        return view.background(.regularMaterial, in: .containerRelative)
+            .padding()
     }
 }
--- a/PlayCover/Utils/Extensions/ViewExtensions.swift
+++ b/PlayCover/Utils/Extensions/ViewExtensions.swift
@@ -1,0 +1,34 @@
+//
+//  URLExtensions.swift
+//  PlayCover
+//
+
+import SwiftUI
+
+extension View {
+    @ViewBuilder
+    func toastOverlay<Content: View>(
+        @ViewBuilder content: () -> Content
+    ) -> some View {
+        if #available(macOS 26.0, *) {
+            self.safeAreaBar(edge: .bottom) {
+                content()
+            }
+        } else {
+            self.overlay {
+                content()
+            }
+        }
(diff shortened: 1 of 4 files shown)


PlayCover/Views/MainView.swift
```diff
@@ -175,6 +181,23 @@ struct MainView: View {
     }
 }
 
+extension View {
```
TheMoonThatRises: Can you move both `toastOverlay` and `toastBackground` to a separate View extension file?

TheMoonThatRises (changes requested):

PlayCover/Views/MainView.swift
Stv-X: Changes have been made as requested. Please take another look.

PlayCover/Utils/Extensions/ViewExtensions.swift
```diff
@@ -0,0 +1,34 @@
...
+
+import SwiftUI
+
+extension View {
+    @ViewBuilder
+    func toastOverlay<Content: View>(
+        @ViewBuilder content: () -> Content
+    ) -> some View {
+        if #available(macOS 26.0, *) {
+            self.safeAreaBar(edge: .bottom) {
```
TheMoonThatRises: Is the `safeAreaBar` API available pre MacOS 26? If it isn't this may not compile on earlier versions of MacOS, and so this may need to be put in a macro instead of an if-statement. Same with the if-statement below with `glassEffect`.

Stv-X: Thanks for the catch! I’ve updated the implementation using `#if compiler(>=6.2)` (which is the earliest Swift Compiler version bundled with the macOS 26 SDK, according to [official information](https://developer.apple.com/support/xcode)) to wrap these calls. This ensures that:

1. **At build time**: The compiler only attempts to compile these new APIs if the SDK supports them.
2. **At runtime**: It still uses `if #available` to ensure backward compatibility for users on older macOS versions.

Let me know if this approach aligns with the project's standards!

TheMoonThatRises:
Currently it seems that the padding on the right side of the toast is flush with the edge of the app instead of having the same padding as the bottom. This behavior I encounter when compiling, running, and attempting to download an app seems contrary to the images provided in your preview. Can you fix that as currently it makes the app look unpolished?

Stv-X:
> Currently it seems that the padding on the right side of the toast is flush with the edge of the app instead of having the same padding as the bottom. This behavior I encounter when compiling, running, and attempting to download an app seems contrary to the images provided in your preview. Can you fix that as currently it makes the app look unpolished?

Sorry for the oversight. Similar to the previous issue, I didn't thoroughly test the UI on SDKs prior to macOS 26, which led to the discrepancy you observed. >
The issue was caused by the margin adjustments intended for macOS 26's new design leaking into older versions. I have wrapped the specific logic in compiler checks to ensure the toast remains properly inset regardless of the environment. The latest commit includes this fix.

TheMoonThatRises (approved):

PR #2122, opened 2026-04-13
Replace global `npm` install with `npx` runtime

TheMoonThatRises:
Note that `graphicsmagick` and `imagemagick` dependencies for `create-dmg` was removed in a new release. Updates for `checkout` and `upload-artifact` versions across actions was also committed.


JoseMoreville:
Dependency updates done in: https://github.com/sindresorhus/create-dmg/pull/100

JoseMoreville (approved):

Issue #2106, opened 2026-03-24
[Bug]: PlayCover crashes when installing  洛克王国：世界


Leon-Liu228:
### Describe the bug

Hi, I'm trying to install an app into PlayCover, and when it almost finishes, "PlayCover quit unexpectedly". Here is the crash log (Show Details). Please help, many thx!

### Enviroment

[...]

PR #2126, opened 2026-04-18
Fix crash on custom user directory path

Max-Sum:
If user changed the user directory, PlayTool will fail to read files in PlayCover container, including keymap configs and other stuff.

This may have caused issues like #2106.
This patch fixes it.

### DIFF
--- a/PlayCover/Model/PlayRules.swift
+++ b/PlayCover/Model/PlayRules.swift
@@ -14,8 +14,12 @@ struct PlayRules: Decodable {
     var bypass: [String]?
 
     public static func buildRules(rules: [String], bundleID: String) -> [String] {
+        let homeDirectory = FileManager.default.homeDirectoryForCurrentUser
         var result: [String] = []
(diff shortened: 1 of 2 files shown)


PlayCover/Model/PlayRules.swift
```diff
@@ -14,8 +14,13 @@ struct PlayRules: Decodable {
     var bypass: [String]?
 
     public static func buildRules(rules: [String], bundleID: String) -> [String] {
+        let homeDirectory = FileManager.default.homeDirectoryForCurrentUser
         var result: [String] = []
-        let template = RuleTemplate(data: ["NSUserName": "\(NSUserName())", "BundleID": bundleID])
+        let template = RuleTemplate(data: [
+            "NSUserName": "\(NSUserName())",
```
TheMoonThatRises: Can `NSUserName` be safely removed from the template since it is no longer utilized in the default rules?

TheMoonThatRises (approved):

PR #2185, opened 2026-08-12
fix: write a real Info.plist into the launch alias (macOS 27 -54 permErr)

### LINKED ISSUE #2184: [Bug]:macOS 27: Click launch game pop-up "does not have permission to open "(null)""
### Describe the bug

Under macOS 27 beta 5, clicking the launch button inside PlayCover GUI triggers a system alert:
"The application “PlayCover” does not have permission to open "(null)"".
Occasionally the popup incorrectly references Siri.

Game executable works perfectly when started directly from terminal.
Only GUI launch path fails.

This issue exists both on official nightly build AND viatearz community fork.

### Steps to reproduce

1. Install PlayCover (official nightly / viatearz fork)
2. Import mobile game IPA (Tested: Arknights: Endfield, Zenless Zone Zero)
3. Tap launch button inside PlayCover UI
4. Null permission alert pops up, game fails to launch
5.Workaround: Launch the game executable manually in terminal, bypassing PlayCover GUI startup logic.
Example path for Arknights: Endfield

### Expected behaviour

The game should launch normally after clicking the button without any null permission system alert.

### Crash log

```shell
No full application crash. Only system permission alert pops up, PlayCover itself remains running.
Check Console logs shows the system blocks NSWorkspace open request with null argument.
```

### What version of PlayCover are you using?

Nightly/beta (please specify build number)

### What version of macOS are you using?

macOS beta (please state the specific version)

### Issue Language

- [x] Yes my issue is written in English

Seraphim0916:
Fixes #2184.

### Problem

On macOS 27, clicking the launch button shows *"The application 'PlayCover' does not have permission to open '(null)'"* and the app never starts. Double-clicking the alias in `~/Applications/PlayCover/` from Finder fails the same way.

`runAppExec()` opens `aliasURL` — `~/Applications/PlayCover/<Name>.app` — which `createAlias()` builds as a directory populated with one symlink per top-level item of the real bundle. LaunchServices on macOS 27 refuses to open that: `"(null)"` is it failing to resolve an app identity for the path.

The symlinked `Info.plist` is what it rejects.

### Change

Copy `Info.plist` into the alias as a real file and leave every other item a symlink.

This keeps the alias directory itself as the launched bundle, which matters: it is what makes the Dock, the app switcher and Force Quit show the app's PlayCover name. Replacing the whole alias with a single symlink to the bundle also fixes the launch, but the name then degrades to the bundle identifier (`com.example.game` instead of the display name), so this patch keeps the existing alias shape.

The copy cannot go stale — `PlayApp.init` calls `removeAlias()` then `createAlias()`, so the alias is rebuilt from the current bundle on every launch.

### Evidence

macOS 27.0 (26A5406e), same app, three alias shapes in `~/Applications/`:

```
# every item symlinked — current behaviour
$ open ~/Applications/PlayCover/<Name>.app
_LSOpenURLsWithCompletionHandler() failed with error -54

# Info.plist copied, everything else symlinked — this patch
$ open ~/Applications/Test/<Name>.app
$ echo $?
0                       # LaunchServices reports the app as "<Name>"

# one symlink to the whole bundle
$ open ~/Applications/Test2/<Name>.app
$ echo $?
0                       # but LaunchServices reports it as "com.example.game"
```

`-54` is what surfaces as the alert. From the GUI launch button on nightly 1571:

```
CoreServicesUIAgent  [com.apple.launchservices:default] LAUNCH: Launch failed in CSUI with error
Error Domain=NSOSStatusErrorDomain Code=-54 "permErr: permissions error (on file open)"
UserInfo={_LSLine=4288, _LSFunction=_LSOpenStuffCallLocal, _LSFile=LSOpenCore.mm, _LSErrorMessage=permErr}
```

### Verification

Built this branch (`xcodebuild -scheme PlayCover -configuration Release`, Xcode 27.0/27A5228h) and ran it on macOS 27.0:

- the four existing aliases were rebuilt on first launch, each with a real `Info.plist`;
- the launch button starts apps again — no `permErr` in the log;
- `lsappinfo` reports each launched app under its PlayCover name.

### DIFF
--- a/PlayCover/Utils/Extensions/PlayAppExtensions.swift
+++ b/PlayCover/Utils/Extensions/PlayAppExtensions.swift
@@ -57,9 +57,18 @@ extension PlayApp {
                                                     withIntermediateDirectories: true,
                                                     attributes: nil)
             url.enumerateContents(options: [.skipsSubdirectoryDescendants]) { ctxUrl, _ in
-                try FileManager.default.createSymbolicLink(
-                    at: self.aliasURL.appendingPathComponent(ctxUrl.lastPathComponent),
-                    withDestinationURL: ctxUrl)
+                let aliasItem = self.aliasURL.appendingPathComponent(ctxUrl.lastPathComponent)
+
+                // LaunchServices on macOS 27 will not open the alias while Info.plist is a
+                // symlink: it fails with -54 (permErr) and the user sees "does not have
+                // permission to open (null)". A real copy is enough, and it stays current
+                // because the alias is rebuilt on every PlayApp init.
+                if ctxUrl.lastPathComponent == "Info.plist" {
+                    try FileManager.default.copyItem(at: ctxUrl, to: aliasItem)
+                } else {
+                    try FileManager.default.createSymbolicLink(at: aliasItem,
+                                                               withDestinationURL: ctxUrl)
+                }
             }
         } catch {
             Log.shared.log(error.localizedDescription)


Seraphim0916:
Heads up: I force-pushed a different fix than the one I opened this PR with, and the title and description now describe the new one.

The first version replaced the whole alias with a single symlink to the bundle. It launches fine, but it is wrong: the alias directory *is* the bundle LaunchServices launches, so pointing at the real bundle instead makes the Dock, the app switcher and Force Quit fall back to the bundle identifier (`com.example.game`) instead of the app's PlayCover name. My tester spotted it in the Dock within minutes — the alias exists to carry that name, and I had missed why.

The version now on the branch keeps the alias shape exactly as it is and only writes `Info.plist` as a real file instead of a symlink. That one file is what LaunchServices rejects; with it copied, the alias opens and the name is preserved.

Sorry for the churn.

TheMoonThatRises:
Is this bugfix for macOS 27 Beta 5? I have not experienced this issue on Beta 4.

Seraphim0916:
Yes — macOS 27 beta 5. The machine is `26A5406e`, which is developer beta 5 / public beta 3, seeded on Aug 10. The reporter on #2184 also states beta 5, so your beta 4 experience is not a contradiction: something in LaunchServices appears to have tightened between the two.

I cannot pin the exact build where it changed. Both Macs I have on 27 are already on beta 5, so I have no way to test beta 4, or to check whether this survives into the release build.

Worth noting the patch is a no-op wherever the current code already works: it writes `Info.plist` as a real file instead of a symlink, and the alias is rebuilt on every `PlayApp.init`, so on a system that accepts the symlinked form nothing changes except one small file copy. If you would rather wait until a later beta shows whether Apple keeps this behaviour, that is entirely reasonable — the bug is real on beta 5 either way.

MWAITINGU:
Thanks for the fix.No wonder I kept getting errors when trying to compile and run the game in Xcode. I have no idea why Apple is suddenly tightening these rules.:)

TheMoonThatRises (approved):

PR #2063, opened 2026-01-22
Update: add M5 iPad Pro devices

yuemingruoan:

PR #2187, opened 2026-08-12
chore: add M5 iPad Pro and iPhone 17 Pro Max

RyouDYFZ:
Make M5 iPad Pro and iPhone 17 Pro Max available, referenced #2063, co-authored with @yuemingruoan.


RyouDYFZ:
Also related board ID modification in https://github.com/PlayCover/PlayTools/pull/231

PR #2195, opened 2026-08-17
perf: stabilize app library refresh

EmilyTsum:
## Summary
- move directory and Info.plist scanning off the main actor
- cancel stale refresh work
- reconcile the bundle-ID cache atomically
- preserve existing PlayApp instances when the installed bundle is unchanged

## Testing
- SwiftLint --strict

### DIFF
--- a/PlayCover/ViewModel/AppsVM.swift
+++ b/PlayCover/ViewModel/AppsVM.swift
@@ -5,10 +5,80 @@
 
 import Foundation
 
-class AppsVM: ObservableObject {
+private struct AppLibraryEntry: Sendable {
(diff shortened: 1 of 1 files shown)


PlayCover/ViewModel/AppsVM.swift
```diff
@@ -5,10 +5,80 @@
 
 import Foundation
 
-class AppsVM: ObservableObject {
+private struct AppLibraryEntry: Sendable {
```
TheMoonThatRises, line 8: Can you move these new `struct` and `enum`s to a seperate file?

PR #2208, opened 2026-08-27
feat: user custom plugins

ohaiibuzzle:
Allow the user to load their own dylibs (which can contain fixes for apps) as needed by themselves. Needs the accompanying PlayTools PR (https://github.com/PlayCover/PlayTools/pull/236).

Also allows for us to later be able to chop PlayTools into pieces so it isn't a giant blob.


JoseMoreville (approved):
Looks good to me and worked as intended

Issue #836, opened 2023-03-03
[Feature]:  hey, can you support  open multiple applications ？ thanks


liuqijie6:
hey, can you support  open multiple applications ？ thanks

Like the Android emulator, it supports multiple applications

### Issue Language

- [X] Yes my issue is written in English

ohaiibuzzle:
Multiple applications can be opened at the same time. 
Multiple instances of one application **will not be supported**. This is due to various technical reasons, as well as preventing PlayCover from being used as for bots.

PR #2219, opened 2026-09-20
feat: duplicate an app so two copies can be played at once

### LINKED ISSUE #2173: [Feature]: Running two instances of one game
大佬们，能否支持同一个游戏同时开启 2 个实例，

比如开两个原神。

Hey everyone, is it possible to support running two instances of the same game simultaneously—for example, having two copies of *Genshin Impact* open at once?

anandghegde:
Closes #2173.

Adds **Duplicate App** to an app's context menu in the library. It makes a second, independent copy of an installed app, so both can be played at the same time, each signed into its own account.

### Why this is small

Everything PlayCover keeps for an app is stored under that app's bundle identifier — the container, `App Settings/<id>.plist`, `Keymapping/<id>/`, `Entitlements/<id>.plist`, the PlayChain. So a copy only needs an identifier of its own to get all of those separately, and a duplicate is just the same bundle installed under a new identifier and a new name, signed again.

Nothing about launching, containers, keychains or PlayTools had to change.

### What it does

- **Names.** `com.foo.game` → `com.foo.game.copy2`, and `Game` → `Game 2`. Both are checked against what is installed before being offered, and the name is checked again after the user edits it: the alias in `~/Applications/PlayCover` is named after the app, so two apps sharing a name would share one alias and the second would silently launch the first. Duplicating a duplicate numbers from the original rather than growing `.copy2.copy2`.
- **Disk.** The bundle is cloned with `clonefile(2)` rather than copied. A 60 GB game duplicates instantly and costs no extra disk until one of the two is written to. Volumes that cannot clone fall back to `FileManager.copyItem`.
- **Bundle contents.** App extensions under `PlugIns/` move underneath the copy's identifier, since they have to stay prefixed by their host app's. Frameworks are deliberately left alone — nothing requires renaming them, and apps do look their own frameworks up by identifier.
- **Settings and keymaps** are carried over and retargeted, so the copy starts out playing the same way. `KeymapConfig` stores absolute URLs, so the keymaps are adopted one at a time and the default-keymap pointer is rewritten, rather than the directory being copied wholesale.
- **Entitlements** dumped at install time are carried across so the copy signs the same way.
- **App data is not carried over.** The reason to want a second copy is a second account, and a copy of a container that is already signed in is not that.
- A copy that fails halfway through is deleted rather than left in the library as a broken app.
- Progress and the finished toast reuse `InstallVM`, so there is no new UI plumbing — one extra case in `InstallStepsNative`.

### Testing

Builds clean against `develop`, and SwiftLint `--strict` reports nothing new (only the nine pre-existing `legacy_swiftui_aspect_ratio` warnings in untouched files).

I don't have a real IPA on this machine to try this against, so rather than claim more than I checked, I put the actual `AppDuplicator` and `Keymapping.adoptKeymaps` into a throwaway harness with a synthetic wrapped app — `Info.plist`, an executable, an `.appex` of its own, a third-party `.appex`, a framework, two keymaps with the non-default one marked as default, settings and entitlements — and asserted the behaviour above: identifiers and names, that the app it was copied from is left untouched, that the third-party extension and the framework keep their identifiers, that the copy's keymaps are retargeted and its default-keymap pointer lands in its own directory, that settings are carried over and retargeted, that writing to the clone does not write through to the original, that numbering a copy-of-a-copy and a taken name both behave, that copying over an installed app is refused, and that a copy that fails is cleaned up. Happy to bring that in as a test target if you'd want it — the project doesn't have one today, so I left it out.

What I have **not** been able to do is launch two copies of a real game side by side. If someone with a game installed can try it, that's the part worth a second pair of eyes.


TheMoonThatRises:
Directly supporting multiple instances of the same app will not be supported (see https://github.com/PlayCover/PlayCover/issues/836). There are workarounds to manually do launch the same app in multiple instances.

